PR #8, opened 2022-05-03
jaq requires rust >= 1.59

sanpii:
https://github.com/01mf02/jaq/blob/main/jaq/src/main.rs#L249


01mf02:
Thank you for your contribution, I did not know about `rust-version`!

PR #15, opened 2022-06-22
docs: add homebrew installation

chenrui333:
add [homebrew installation](https://formulae.brew.sh/formula/jaq#default)


chenrui333:
> Ah, now I understand why you were interested in having the releases tagged! :) Thank you for making the homebrew formula & enjoy using jaq. ;)

yeah, indeed :D

PR #27, opened 2022-07-19
Fix --join-output (-j) to output raw strings

itchyny:
This PR fixes `--join-output (-j)` option to behave more compatible with jq, to emit strings in raw mode.
```
 $ jq -n -j 'null,true,"hello, world",42,[]' 
nulltruehello, world42[]

 $ jaq -n -j 'null,true,"hello, world",42,[]' 
nulltrue"hello, world"42[]
```

### DIFF
--- a/jaq/src/main.rs
+++ b/jaq/src/main.rs
@@ -299,7 +299,7 @@ fn run_and_print(
 
     let last = run(filter, ctx, iter, |output| {
         match output {
-            Val::Str(s) if cli.raw_output => print!("{}", s),
+            Val::Str(s) if cli.raw_output || cli.join_output => print!("{}", s),
             _ => {
                 // this looks ugly, but it is hard to abstract over the `Formatter` because
                 // we cannot create a `Box<dyn Formatter>` because


01mf02:
I have thought about this a bit. What are the pros and cons of this change?

Pros:

* Increased compatibility with jq

Cons:

* Less flexible than current solution: jaq can currently simulate jq's `-j` flag with `-jr`, but jq cannot simulate jaq's `-j` flag.
* More complicated code: Not much, but still.
* More complex documentation: Currently, jaq's documentation of `-j` is simply: "Do not print a newline after each value". Adding something like "... and write strings without escaping them with quotes" more than doubles the length of the documentation.

I suppose that a user who knows jq well enough to know about `-j` will probably also know it well enough to see quite quickly what to do when jaq does not escape strings, namely just add `-r`. Therefore, I believe it would be a better solution to keep jaq's current behaviour, but to document the difference in the `Cli`. What do you think?

itchyny:
I'm fine with the position of jaq that it is a enhanced and well-behaved successor, not a drop-in replacement of jq.

01mf02:
@itchyny, I'm happy to hear that. So I documented the behaviour of `-j` in 561aab9.

PR #29, opened 2022-08-05
Add while to standard library

capezotte:
This adds the `while` command from jq's standard library.

```
$ ./target/debug/jaq '[ while(. < 100; . * 2) ]' <<< 1
[
  1,
  2,
  4,
  8,
  16,
  32,
  64
]
```

### DIFF
--- a/jaq-std/src/std.jq
+++ b/jaq-std/src/std.jq
@@ -52,6 +52,7 @@ def range(x): range(0; x);
 def repeat(g): [g] | recurse(.) | .[];
 def recurse: recurse(.[]?);
 def recurse(f; cond): recurse(f | select(cond));
+def while(cond; update): recurse(if cond then update else empty end; cond);
 
 # Iterators
 def map(f): [.[] | f];
--- a/jaq-std/tests/std.rs
+++ b/jaq-std/tests/std.rs
@@ -212,3 +212,10 @@ fn typ() {
     give(json!(true), "type", json!("boolean"));
     give(json!(null), "type", json!("null"));
 }
+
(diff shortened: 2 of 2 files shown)


capezotte:
Fixed a typo that caused one more object to appear in comparison to jq, sorry.

01mf02:
Thanks a lot for your contribution!
Would you mind adding some tests for your new filters to `jaq-std/tests/std.rs`?

capezotte:
Done.

01mf02:
I found a difference between your implementation and jq's one: `until(.; (1, 2))` with input `null` yields `1 2` with jq's builtin `until`, whereas with your implementation of `until`, it yields just `2`.

01mf02:
Thanks for your tests, by the way.
Just to be clear: If you find a nice (read: jq-compliant and performant) way to implement `until`, then I'll be happy to merge your PR. :) But if not, then I'll be also fine with only merging your `while` implementation.

capezotte:
So far I don't see a way to implement `until` with just the existing functions without failing tests with commas. Guess I'll drop it.

01mf02:
Thanks a lot!
I felt inspired by your effort and implemented `until` in 9cda5b086507014d5d48074749d156bda71400f5 as core filter --- it turned out that this was just a slight generalisation of the `recurse` filter.
If you still have your tests for `until`, I would be happy to incorporate them, perhaps in a new pull request?

01mf02:
Ok, I went forward and incorporated your tests for `until` in c8190bb. Thanks again!

PR #46, opened 2022-10-10
Update Brainfuck example of Fibonacci sequence

itchyny:
I noticed that you copied some Brainfuck codes from my 10 year old repository. That's totally fine, but I found the Brainfuck code can be slightly shorter. I know this is irrelevant improvement to this repository but can't help sending PR. Good luck with the compatibility improvement.


01mf02:
Thanks a lot! Your updated Fibonacci generator is actually a very nice improvement to this repository, because I included it in my benchmarks, and your new version is about 26% faster. :)

PR #59, opened 2023-02-25
Make Filter struct Clone

elkowar:
Hey! I'm planning on using `jaq_core` to have embedded JQ-style syntax in my widget system [eww](https://github.com/elkowar/eww).
As the filters are gonna be ran several times, I'd want to cache the generated filters in between runs. For this, the filter struct needs to be clone!


01mf02:
Sounds good to me! If you hit any further roadblocks during the integration, let me know!

PR #65, opened 2023-03-19
Add group_by filter

Shirtpantswallet:
This is admittedly not the most efficient implementation of `group_by`, due to the redundant calls to `f`, but am hoping it's useful enough to include.


01mf02:
Thank you for your implementation! However, I want jaq to compile on stable Rust, but you use the unstable `slice_group_by` feature. If you can come up with an implementation without unstable features, I'll be happy to include it.

jaq-core/src/val.rs
01mf02, line 1: To avoid calling `f` multiple times on the same value, couldn't we just map all values v to (v | f, v), then sort and group this, and then remove the `v | f` again?

01mf02 (changes requested):
Thanks for your changes. Could you possibly also add some tests for this filter to jaq-core/tests/named.rs?

jaq-core/src/val.rs
```diff
@@ -294,15 +295,18 @@ impl Val {
...
-                    == f(y.clone()).filter_map(|v| v.ok()).collect::<Vec<_>>()
-            })
-            .map(|v| Val::Arr(Rc::new(v.to_vec())))
-            .collect::<Vec<_>>(),
-        )))
+        let mut a = self.sort_by(&f)?.to_arr()?;
+        Ok(Val::Arr(
+            Rc::make_mut(&mut a)
+                .into_iter()
+                .group_by(|x| f((*x).clone()).collect::<Result<Vec<Val>, _>>().unwrap())
```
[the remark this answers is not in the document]
Shirtpantswallet: Given that `sort_by` has already called `f()` on every value and not failed, I believe the `.unwrap()` to be guaranteed to not fail. I will be sure to include a guard instead when I combine the sort and group stages.

jaq-core/src/val.rs
Shirtpantswallet: I believe so! I'll give it another go.

Shirtpantswallet: I re-implemented it as a BTree instead of a sort + group, which yielded a significant speedup. However, I realized that the implementation is actually wrong and will be correcting it.

PR #67, opened 2023-03-20
Unique pr

Shirtpantswallet:
This implementation was lifted directly from jq's [builtin.jq](https://github.com/stedolan/jq/blob/master/src/builtin.jq). Would you like me to add tests for this as well?


01mf02:
Thanks for offering to add tests for this, but I think it's fine like this because you reused precisely the definition present in jq.

01mf02:
@wader, I think that something like `--run-tests` is not really necessary for jaq, because `cargo test` runs all tests anyway; no need to encode tests into the final binary.

wader:
Aha sorry i meant more to be able to run jq test suit to test computability, assuming that is a goal. Think i've also seen ppl use it to unit test jq code, for fq i use a jq implementation of it (uses a eval function) to test some additional standard library jq functions.

01mf02:
By the way, I checked out your work on jqjq ... that's some quite impressive work! It has made me think about how to efficiently implement recursive & nested functions, given that you seem to use these a lot ...

wader:
Thanks! yeap i've started to use nested functions a lot in jq, nice for structure and can be very powerful. I'm not sure how good jqjq is as an example for common jq code :) but maybe nice for profiling?

BTW i did try to run jqjq with jaq some time ago but ran into some issue, i think nested functions and maybe binding arguments were the obvious missing things, maybe some CLI args also? let me know if you want to cooperate on that. I haven't done much rust but would like to learn more.

Also thanks for jaq, been browsing the code from time to time, learned a lot, especially the parse code when i worked on jqjq. Without the references to precedence climbing i'm not sure jqjq would have happen!

01mf02:
Concerning what was missing from jaq to run jqjq, I guess it would be best to look at this once I have implemented recursive/nested functions. I'm progressing steadily, but it is not so straightforward to do this efficiently (and correctly ^^).

I feel humbled that jaq has served as an inspiration for your project, and I am happy to hear that it was useful to you.

wader:
> Concerning what was missing from jaq to run jqjq, I guess it would be best to look at this once I have implemented recursive/nested functions. I'm progressing steadily, but it is not so straightforward to do this efficiently (and correctly ^^).

👍  no hurry! :) i'm keeping an on the jaq development, interesting to follow along, and hopefully i will be able to help.

> I feel humbled that jaq has served as an inspiration for your project, and I am happy to hear that it was useful to you.

Specifically the parsing stuff was very helpful for jqjq, did several attempts at doing various LL/LR parsing in jq but always got stuck with infinite recursion or very complex rule rewrites to get the infix operators to work. Precedence climbing solved all of that and was very quick! ... remember now that i even peeked at how hard it would be to make a jq yacc thingy :)

PR #74, opened 2023-04-22
Add min_by, max_by

kklingenberg:
This PR is a naive attempt to implement the missing `min_by` and `max_by` filters. I'm unsure about the part on `run_if_ok` and checking whether an error happened after reduction, but in the end I decided to imitate the `sort_by` and `group_by` functions on that, since I'm not very sure about how things work in the library, generally.

I was in doubt about sending the PR because this seems like a pair of functions that could have been implemented directly in `std.jq` using a map-reduce approach, but having tried that and benchmarked against gojq, I saw that it had to be provided from Rust to be comparable. In the end, and since I don't have much of a clue on how to make these things fast, I "ported" my jq implementation to Rust :sweat_smile:.

Hopefully this is useful.


01mf02:
Hi @kklingenberg, thanks a lot for your PR!
I discovered a case where your implementation diverges from jq;
`[{a: 1, b: 2}, {a: 1, b: 3}] | max_by(.a) | .b` yields 3 in jq, but 2 in jaq.
Furthermore, it would be good to add a test where the function passed to `max_by` yields multiple outputs.
When you address both these issues, I will be happy to merge your PR.

kklingenberg:
Hello @01mf02 ! I've updated the PR and it should now follow your suggestions. Thank you! And thank you again for building jaq!

01mf02:
I reworked your implementation by making it abort on the first error, instead of continuing until the end of the array and only then yielding the error.
Furthermore, because the tests are now for a native function, I moved most of them to `jaq-core/tests`.
Thanks again for your contribution!

PR #76, opened 2023-05-14
Fix Native filters

passcod:
- Publicly expose `filter::Native` -- otherwise that API just can't be used!
- Re-enable (and fix the types of) the custom filters integration tests -- probably why you didn't catch that this broke during your refactors.
- Derive `Clone` on Definitions again, and add `Default`, to cover the case where there's no variables -- can just be done with `Definitions::new(Vec::new())` though so this is a small convenience only.

### DIFF
--- a/jaq-core/src/lib.rs
+++ b/jaq-core/src/lib.rs
@@ -137,6 +137,7 @@ impl Filter {
 ///
 /// For example, if we define a filter `def map(f): [.[] | f]`,
 /// then the definitions will associate `map/1` to its definition.
+#[derive(Debug, Clone)]
 pub struct Definitions(mir::Defs);
 
 impl Definitions {
--- a/jaq-core/src/lib.rs
+++ b/jaq-core/src/lib.rs
@@ -61,6 +61,7 @@ mod val;
 pub use jaq_parse as parse;
 
 pub use error::Error;
+pub use filter::Native;
 pub use rc_iter::RcIter;
 pub use val::{Val, ValR};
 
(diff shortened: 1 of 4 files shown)


passcod:
Closing as I see you're doing that https://github.com/01mf02/jaq/commit/01135fb5d6fbc2e1ef24d2b687aa23feb4932842

01mf02:
Hi @passcod, sorry for ignoring your PR for such a long time.
In the end, as you have mentioned, I redid this thing myself, because I wanted to have a nice API that exposes as little implementation details about filters as possible. Now, by using the `FilterT` trait, native filters can also call filters that are passed as arguments.

I am interested about one detail. You write in a7157bc6acc32884c40736d9fecd3979f48c2c6a:

> This makes the API tons easier to use as there’s no need to construct
a whole new Definitions each and every time a filter needs calling.

This surprises me, because normally, you should be able to just "reuse" a filter after having called it. Would you like to explain to me your use case?

passcod:
I don't have the code in front on me right now but the filter running method used to take `self`, so it consumed the `Definitions`. As that didn't have Clone it was impossible to use twice (with two different inputs), unless I was using it completely wrong.

---

My usecase is running a series of filters on a filesystem (for example) event, to decide whether to discard it or keep it. There may be thousands of such events coming in per second. Filters have an application-specific library on top of the standard jaq ones, and are provided at the start of the program lifetime and not changed.

01mf02:
> I don't have the code in front on me right now but the filter running method used to take `self`, so it consumed the `Definitions`. As that didn't have Clone it was impossible to use twice (with two different inputs), unless I was using it completely wrong.

Ok. At least with the current API, filters can be run without consuming them, as can be seen in the [jaq CLI](https://github.com/01mf02/jaq/blob/b072c6214adb974c9b235a5cd5faecf857e9bee4/jaq/src/main.rs#L405). If you think that there is still a use case for having `Definitions` implement `Clone`, feel free to open a PR.

PR #78, opened 2023-06-02
docs: Mention unique under core filters

jpmckinney:
```
$ echo '[1,2,5,3,5,3,1,3]' | jaq 'unique'
Error: could not find function
```

Adding to docs to make it easier to confirm that it's not implemented.


jpmckinney:
Ah, I see it's implemented but not in a release yet.

PR #81, opened 2023-06-19
Add keys_unsorted

qubidt:
To match jq, rename `keys` to `keys_unsorted` and implement `keys` as a std filter based on `keys_unsorted`. `keys` now returns the array of keys sorted


qubidt:
Also note that while jq specifies that sorting is only performed if the input is an object. That could be enforced in the filter definition but I opted to keep the definition simple.

01mf02:
That looks good, thanks! Let's keep your simple filter definition, as I suppose that `keys` is used mostly on objects anyway.

PR #83, opened 2023-06-20
Implement gsub(regex; string; flags)

qubidt:
Add missing function: `gsub` with 3 args


01mf02:
Thank you for your addition, especially your illustrative tests!

PR #84, opened 2023-06-20
Imeplement string functions

qubidt:
Adds all (AFAICT) of the missing string functions:

* `startswith`
* `endswith`
* `ltrimstr`
* `rtrimstr`


01mf02:
Thanks once more!

PR #85, opened 2023-06-20
Implement date functions

baod-rate:
Implement date functions:

* `now`
* `fromdateiso8601`
* `todateiso8601`


qubidt:
`todateiso8601` produces different formats depending on the input type:
* `YYYY-MM-DDThh:mm:ss±hh:mm` for ints 
* `YYYY-MM-DDThh:mm:ss.sssssssss±hh:mm` for floats

I wanted to do the similar for `fromdateiso8601` but wasn't happy with how bloated it made the function. Implementation suggestions would be appreciated

The other date functions (i.e. strftime/strptime) don't have equivalents in the time crate. Could maybe use https://github.com/MiSawa/time-fmt/

01mf02:
> I wanted to do the similar for `fromdateiso8601` but wasn't happy with how bloated it made the function. Implementation suggestions would be appreciated

To help you with this, would you like to elaborate how you were trying to approach the problem?

> The other date functions (i.e. strftime/strptime) don't have equivalents in the time crate. Could maybe use https://github.com/MiSawa/time-fmt/

I'd say let's go for now only for the ISO format, and postpone the hard case if there is sufficient demand later. :)

qubidt:
> To help you with this, would you like to elaborate how you were trying to approach the problem?

I don't remember exactly what I did, but it was pretty involved. Looking at it with fresh eyes, the problem seems a lot simpler. 😅 I added the new implementation in 7bc1b54a56d864345860087846558fdf1b47454f. tell me what you think

qubidt:
Looks like tests are failing because I added `time` as an optional dependency without properly feature-gating all the code, but I got a bit stuck on the best way to conditionally chain the datetime filters to the core filters: 

https://github.com/01mf02/jaq/blob/5184281304ad4c2ad92f2b311ddf274475401008/jaq-core/src/core.rs#L12-L21

Should I just add `time` a non-optional dependency? Or should I go ahead and move all the code that depends on it behind feature-gates?

01mf02:
You can go ahead and make `time` a non-optional dependency.

01mf02:
> I don't remember exactly what I did, but it was pretty involved. Looking at it with fresh eyes, the problem seems a lot simpler. sweat_smile I added the new implementation in [7bc1b54](https://github.com/01mf02/jaq/commit/7bc1b54a56d864345860087846558fdf1b47454f). tell me what you think

That looks nice now that the functions round-trip.

baod-rate:
Just now realized that `SystemTime` requires `std`, which means `now` is a no-go. How do you think this case should be handled?

01mf02:
Don't worry about `std`. I have a plan anyway to change the core filter implementation, and this will address the `std` issue.

PR #86, opened 2023-06-26
Implemented str * int operations

nouritsu:
Feature Addition - 
`str * int` will result in a string that is repeated `int` times
If int is negative or zero, resulting value will be `null`.

### DIFF
--- a/jaq-core/src/val.rs
+++ b/jaq-core/src/val.rs
@@ -589,8 +589,14 @@ impl core::ops::Mul for Val {
             (Int(x), Int(y)) => Ok(Int(x * y)),
             (Float(f), Int(i)) | (Int(i), Float(f)) => Ok(Float(f * i as f64)),
             (Float(x), Float(y)) => Ok(Float(x * y)),
-            (Num(n), r) => Self::from_dec_str(&n) * r,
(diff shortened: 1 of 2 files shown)


nouritsu:
Question: Where do I add tests? I could not find a test for multiplication and division in the jaq-core/tests/ directory.

However, I have tested the implementation on the command line and it works as expected -
```
> jaq -n '5 * \"hello\"' 
"hellohellohellohellohello"
```
```
> jaq -n '\"hello\" * 5' 
"hellohellohellohellohello"
```
```
> jaq -n '\"hello\" * -5'
null
```
```
> jaq -n '\"hello\" * 0'
null
```
This indicates that the commutative property is satisfied and the results obtained are the exact same as on the latest version of jq.
```
> jq -n '5 * \"hello\"' 
"hellohellohellohellohello"
```
```
> jq -n '\"hello\" * 5' 
"hellohellohellohellohello"
```
```
> jq -n '\"hello\" * -5'
null
```
```
> jq -n '\"hello\" * 0'
null
```

nouritsu:
Tests have been added, this is ready to merge.

01mf02:
Thanks a lot!

PR #91, opened 2023-07-07
Implement --nul-output

baod-rate:
For compatibility with jq's upcoming release.

Introduced to jq in https://github.com/jqlang/jq/pull/1990

Note that JSON can contain NUL characters (e.g. `"\u0000"`), so it is not possible to distinguish separate values in every case, even when using this flag


01mf02:
Thanks!

pkoppstein:
@baod-rate - Please see https://github.com/01mf02/jaq/issues/101

PR #92, opened 2023-07-07
Match jq exit code for no output with --exit-status

baod-rate:
For compatibility with jq's upcoming (1.7?) release.

Introduced into jq with https://github.com/jqlang/jq/pull/2654, when using --exit-status, parse errors now cause jq to exit with status code 5 (`JQ_ERROR_UNKNOWN`) instead of 4 (`JQ_OK_NO_OUTPUT`). The motivation is to be able to distinguish the no-output case from parsing errors.

Mirroring this, 5 will be used for both `Error::Jaq` and `Error::Parse`, i.e. both filter parsing errors and JSON parsing errors.

### DIFF
--- a/jaq/src/main.rs
+++ b/jaq/src/main.rs
@@ -179,12 +179,14 @@ fn real_main() -> Result<ExitCode, Error> {
         last
     };
 
-    Ok(if cli.exit_status {
-        // return exit code 4 if no value is output
-        ExitCode::from(last.map(|b| (!b).into()).unwrap_or(4))
+    if cli.exit_status {
+        last.map_or_else(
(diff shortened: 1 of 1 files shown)


baod-rate:
I also added the non-zero exit conditions to the `Error` enum, hoping to centralize the exit codes, and to make the exit conditional more readable. Please tell me if that's unidiomatic or otherwise subpar style though.

01mf02:
> I also added the non-zero exit conditions to the `Error` enum, hoping to centralize the exit codes, and to make the exit conditional more readable. Please tell me if that's unidiomatic or otherwise subpar style though.

I like your change. :)

Thanks again!

PR #94, opened 2023-07-15
Embed the current abbreviated commit SHA into the version

### LINKED ISSUE #40: jaq --version
`jaq --version` currently only gives the "release" version, so it is not possible to use the output of this flag to distinguish between various git commits.  

If it is too much trouble to give at least some part of the git sha, then it would still be helpful to distinguish the "official release" from a subsequent version.

kklingenberg:
Closes #40

### DIFF
--- a/jaq/build.rs
+++ b/jaq/build.rs
@@ -1,0 +1,13 @@
+use std::process::Command;
+
+fn main() {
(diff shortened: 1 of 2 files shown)


01mf02:
Thanks a lot!

PR #96, opened 2023-07-19
Add math filters from libm

### LINKED ISSUE #88: libm's functions
Rust's [libm](https://docs.rs/libm/latest/libm/) provides many of the functions described in [jq's manual](https://jqlang.github.io/jq/manual/#Math), with some exceptions (e.g. `gamma`). However, the manual states that availability depends on the system, since they're linked to the C standard library:

> Unavailable math functions will be defined but will raise an error.

Should jaq do the same, always failing on functions that aren't defined in libm?

Also, and since including these would add another dependency, should they be feature-gated in jaq?

kklingenberg:
This PR adds the range of functions from [libm](https://github.com/rust-lang/libm) as filters, with the exception of `round`, `floor` and `ceil`.

- [x] Port 1-ary functions as 0-ary filters
- [x] Port 2-ary functions as 2-ary filters that ignore `.`
- [x] Port 3-ary functions as 3-ary filters that ignore `.`
- [x] Add filters specified in jq's source (or gojq's) that aren't provided as-is in libm.
- [x] Add tests.

Closes #88


01mf02:
That looks very nice already! Great job!

I would move the macros into a separate file (`math.rs`?) and keep only `const MATH` inside `lib.rs`.
Furthermore, in `math_2_ary`, you can use `xs.cartesian(ys, cv)` to create `(x, y)` pairs efficiently.

kklingenberg:
Thank you! I have many doubts still, and probably should turn them into a checklist:

- How do I test these? Testing the numerical identities seems wrong, since they're tested extensively in libm's crate. And rustc already tests for type safety.
- As stated in the linked issue, there's many functions absent from libm that are listed in jq's manual, **and** there's functions i libm not mentioned in the manual. Should I forgo following the manual (i.e. the intersection of both) in favour of sticking to libm's catalog?
- Some functions take an integer parameter. Should I use the existing error `Int(Val)` for those? Or maybe extend the new error value into something like `MathFn(String, Val, String)` mapping to the message `cannot apply {fun} to {v} (value should be of {t} type)`?

kklingenberg:
Another doubt that surfaced given your suggestion of using `cartesian()`: I noticed the note in the README about consistency in cartesian products, so is it OK to diverge from jq and gojq in these as well? For example:

```text
$ jq -n 'fma(2,1;3,4;4,5)'
10
7
12
8
11
8
13
9
```

Vs

```text
$ jaq -n 'fma(2,1;3,4;4,5)'
10.0
11.0
12.0
13.0
7.0
8.0
8.0
9.0
```

01mf02:
> How do I test these? Testing the numerical invariants seems wrong, since they're tested extensively in libm's crate. And rustc already tests for type safety.

I would write tests for one function taking no arguments (e.g. `sin`), taking two arguments (e.g. `pow`), and taking three arguments `fma`, especially testing the case when arguments produce multiple outputs.

> As stated in the linked issue, there's many functions absent from libm that are listed in jq's manual, **and** there's functions in libm not mentioned in the manual. Should I forgo following the manual (i.e. the intersection of both) in favour of sticking to libm's catalog? EDIT: As a reference we could take gojq's definitions table [here](https://github.com/itchyny/gojq/blob/d579009515e2d20a7227beb722cc784d6be15910/func.go#L116-L176C48). The author filled the holes with custom implementations for each function missing from go's math module. Perhaps it's right to fill rust's holes too?

I would implement in jaq of course the intersection of libm and jq, then try to implement the remaining things as definition in `jaq-std`. For example, `drem` should be easy to implement by definition from `remainder` and `copysign`, which you both already exported from libm to jaq. I would also export `ilogb` from libm, because that way, you can implement `logb` and `significand` by definition. `pow10` by definition too.

Oh, and you exported `scalb` and `scalbln` to the same libm function; is that intended?

> Some functions take an integer parameter. Should I use the existing error `Int(Val)` for those? Or maybe extend the new error value into something like `MathFn(String, Val, String)` mapping to the message `cannot apply {fun} to {v} (value should be of {t} type)`?

I would indeed use `Int(Val)` for this. And for the math functions, I would analogously to `Int(Val)` introduce an error variant `Float(Val)` in `jaq-core` and use that instead of `MathFn`. The idea behind this is that in the long run, I would like `jaq-core` to not contain error variants that are only produced in `jaq-native`. (`Regex`, `Round`, etc. are also candidates that I would like to see removed.)
I know that this decreases the quality of the errors, because we then will not know which math function has thrown the error, but then, the error quality was never very high to begin with and could be increased far more by tracking where exactly a filter was called. But that will be future work ...

01mf02:
> Another doubt that surfaced given your suggestion of using `cartesian()`: I noticed the note in the README about consistency in cartesian products, so is it OK to diverge from jq and gojq in these as well?

Yes --- my intention is that the output should be:

~~~
$ jq -n '(2,1) as $x | (3,4) as $y | (4,5) as $z | fma($x;$y;$z)'
10
11
12
13
7
8
8
9
~~~

Which your current implementation seems to satisfy. :)

jaq-native/src/lib.rs
```diff
@@ -260,6 +269,163 @@ fn now() -> Result<f64, Error> {
...
+    math::math_0_ary!("cos", libm::cos, math::as_float, Val::Float),
+    math::math_0_ary!("cosh", libm::cosh, math::as_float, Val::Float),
+    math::math_0_ary!("erf", libm::erf, math::as_float, Val::Float),
+    math::math_0_ary!("erfc", libm::erfc, math::as_float, Val::Float),
+    math::math_0_ary!("exp", libm::exp, math::as_float, Val::Float),
+    math::math_0_ary!("exp10", libm::exp10, math::as_float, Val::Float),
+    math::math_0_ary!("exp2", libm::exp2, math::as_float, Val::Float),
+    math::math_0_ary!("expm1", libm::expm1, math::as_float, Val::Float),
+    math::math_0_ary!("fabs", libm::fabs, math::as_float, Val::Float),
+    math::math_0_ary!("frexp", libm::frexp, math::as_float, |(m, e)| Val::Arr(
```
01mf02, line 290: Use `Val::arr` here?

jaq-native/src/lib.rs
```diff
@@ -260,6 +269,163 @@ fn now() -> Result<f64, Error> {
...
+    math::math_0_ary!("trunc", libm::trunc, math::as_float, Val::Float),
+    math::math_0_ary!("y0", libm::y0, math::as_float, Val::Float),
+    math::math_0_ary!("y1", libm::y1, math::as_float, Val::Float),
+    math::math_2_ary!(
+        "atan2",
+        libm::atan2,
+        math::as_float,
+        math::as_float,
+        Val::Float
+    ),
```
01mf02, line 324: To make this and following things more compact, I would make some helper macros for the common use case float -> float -> float, int -> float -> float, and float-> int -> float.
Something like `math::fff`, `math::iff`, and `math::fif`.

01mf02: That saves space and makes it much easier to spot differences/typos.

jaq-native/src/math.rs
```diff
@@ -0,0 +1,108 @@
...
+        Val::Num(ref n) => n.parse().or(Err(Error::MathFn(String::from(fun), v))),
+        _ => Err(Error::MathFn(String::from(fun), v)),
+    }
+}
+
+/// Use a value as an integer to be given as an argument to a libm
+/// function.
+pub fn as_int(v: Val, fun: &str) -> Result<i32, Error> {
+    match v {
+        Val::Int(n) => Ok(n as i32),
```
01mf02, line 19: This looks dangerous -- it looks like it could lose information if `isize` is 64 bits.
Better do a `n.try_into()` and fail with error if it fails.

jaq-native/src/lib.rs
```diff
@@ -260,6 +269,163 @@ fn now() -> Result<f64, Error> {
...
+    math::math_0_ary!("gamma", libm::tgamma, math::as_float, Val::Float),
+    math::math_0_ary!("j0", libm::j0, math::as_float, Val::Float),
+    math::math_0_ary!("j1", libm::j1, math::as_float, Val::Float),
+    math::math_0_ary!("lgamma", libm::lgamma, math::as_float, Val::Float),
+    math::math_0_ary!("log", libm::log, math::as_float, Val::Float),
+    math::math_0_ary!("log10", libm::log10, math::as_float, Val::Float),
+    math::math_0_ary!("log1p", libm::log1p, math::as_float, Val::Float),
+    math::math_0_ary!("log2", libm::log2, math::as_float, Val::Float),
+    math::math_0_ary!("logb", math::logb, math::as_float, Val::Float),
+    math::math_0_ary!("modf", libm::modf, math::as_float, |(f, r)| Val::Arr(
```
01mf02, line 302: I would also test `modf`, because its output is quite different from all other functions.

jaq-native/src/lib.rs
01mf02: And it would probably be good to test this one too.

jaq-native/src/math.rs
```diff
@@ -0,0 +1,108 @@
...
+/// Use a value as a float to be given as an argument to a libm
+/// function.
+pub fn as_float(v: Val, fun: &str) -> Result<f64, Error> {
+    match v {
+        Val::Int(n) => Ok(n as f64),
+        Val::Float(n) => Ok(n),
+        Val::Num(ref n) => n.parse().or(Err(Error::MathFn(String::from(fun), v))),
+        _ => Err(Error::MathFn(String::from(fun), v)),
+    }
+}
```
01mf02, line 13: This should move into jaq_core::Val (without the `fun: &str` bit).

jaq-native/src/math.rs
```diff
@@ -0,0 +1,108 @@
...
+                        Ok($codomain($f(x, y, z)))
+                    }),
+            )
+        })
+    };
+}
+
+pub(crate) use math_3_ary;
+
+/// Source: https://github.com/itchyny/gojq/blob/main/func.go#L1374
```
01mf02, line 78: References to GitHub should be permalinks, such as <https://github.com/itchyny/gojq/blob/d579009515e2d20a7227beb722cc784d6be15910/func.go#L1374> (you get these by clicking on the line, then selecting "Copy permalink").

01mf02:
And I would save my time to write benchmarks. You have made very thin wrappers around `libm` functions, so I do not see the point of benchmarking these.

kklingenberg:
> Oh, and you exported scalb and scalbln to the same libm function; is that intended?

Many of these functions I had never seen before 😅 and those `scalb` and `scalbln` I just pulled from gojq [here](https://github.com/itchyny/gojq/blob/d579009515e2d20a7227beb722cc784d6be15910/func.go#L1390-L1396) and concluded they were the same as libm's scalbn based on [this](https://pubs.opengroup.org/onlinepubs/009696899/functions/scalb.html) and the final recommendation:

> Applications should use either scalbln(), scalblnf(), or scalblnl() in preference to this function.

However, I don't really know what I'm doing here.

kklingenberg:
So I kind of covered your suggestions, though I was probably sloppy everywhere. Sadly I ran out of time and I won't be able to improve things until next ~wednesday, so feel free to take over if you want to merge this before then.

Thank you for your pointers and the patience :)

jaq-std/src/std.jq
```diff
@@ -22,6 +22,16 @@ def isinfinite: . == infinite or  . == -infinite;
...
 
+# Math
+def logb:
+    if . == 0 then -infinite
+  elif isinfinite then infinite
+  elif isnan then .
+  else ilogb_ end;
+def pow10:       pow(10; .);
+def significand: scalb(.; ilogb_ | -1 * .);
+def drem($l; r): remainder($l; r) as $x | if $x == 0 then copysign($x; $l) else $x end;
```
01mf02, line 33: It seems to me that you don't need to bind $x here?

jaq-std/src/std.jq
```diff
@@ -22,6 +22,16 @@ def isinfinite: . == infinite or  . == -infinite;
 def isfinite:   isnumber and (isinfinite | not);
 def isnormal:   isnumber and ((. == 0 or isnan or isinfinite) | not);
 
+# Math
+def logb:
```
01mf02, line 26: I would create a test for this, for example using the inputs `(nan, infinite, -infinite, -2, -1, 0, 1, 2)`.

jaq-native/src/lib.rs
```diff
@@ -260,6 +269,69 @@ fn now() -> Result<f64, Error> {
...
+    math::f_f!(erf),
+    math::f_f!(erfc),
+    math::f_f!(exp),
+    math::f_f!(exp10),
+    math::f_f!(exp2),
+    math::f_f!(expm1),
+    math::f_f!(fabs),
+    math::f_fi!(frexp),
+    math::f_f!("gamma", tgamma),
+    math::f_i!("ilogb_", ilogb),
```
01mf02, line 292: I would simply export this as `ilogb`.

jaq-native/src/lib.rs
```diff
@@ -260,6 +269,69 @@ fn now() -> Result<f64, Error> {
...
+    math::f_f!(cosh),
+    math::f_f!(erf),
+    math::f_f!(erfc),
+    math::f_f!(exp),
+    math::f_f!(exp10),
+    math::f_f!(exp2),
+    math::f_f!(expm1),
+    math::f_f!(fabs),
+    math::f_fi!(frexp),
+    math::f_f!("gamma", tgamma),
```
01mf02, line 291: How do you know that gamma corresponds to tgamma?

jaq-native/src/lib.rs
```diff
@@ -260,6 +269,69 @@ fn now() -> Result<f64, Error> {
...
+    math::ff_f!(fmod),
+    math::ff_f!(hypot),
+    math::if_f!(jn),
+    math::fi_f!(ldexp),
+    math::ff_f!(nextafter),
+    math::ff_f!("nexttoward", nextafter),
+    math::ff_f!(pow),
+    math::ff_f!(remainder),
+    math::fi_f!("scalb", scalbn),
+    math::fi_f!("scalbln", scalbn),
```
01mf02, line 330: These two are not the same,  consider the following example:

~~~
$ jq -n 'scalb(1.5; 3.5), scalbln(1.5; 3.5)'
null
12
~~~

jaq-std/src/std.jq
```diff
@@ -22,6 +22,16 @@ def isinfinite: . == infinite or  . == -infinite;
...
+    if . == 0 then -infinite
+  elif isinfinite then infinite
+  elif isnan then .
+  else ilogb_ end;
+def pow10:       pow(10; .);
+def significand: scalb(.; ilogb_ | -1 * .);
```
01mf02, line 32: Does this yield the same output as jq on nan, infinity, 0 and so on?
Some tests would be nice here.

01mf02:
Thanks a lot for integrating my suggestions. I think that your work looks much better now!
Take your time with the remaining changes — I broke my shoulder yesterday, so from me, things will not move very fast. :)

jaq-native/src/lib.rs
kklingenberg: That is surprising! Again, consider I just learned about many of these functions so I'm little more than an inefficient search engine at this point. Still, checking the man pages for both you get:

- [scalb](https://man7.org/linux/man-pages/man3/scalb.3.html):  `x * FLT_RADIX ** exp`
- [scalbln](https://man7.org/linux/man-pages/man3/scalbln.3.html): `x * FLT_RADIX ** exp`

My intuition is that one is an upgraded version of the other, with handling for special cases. However the example you gave when plugged into the expression supposed to match their definition gives (in say, plain jq) doesn't strike me as a special case at all:

```text
$ jq -n '1.5 * pow(2; 3.5)'
16.970562748477143
```

which is what gojq yields for `scalb(1.5; 3.5)`.

jaq-native/src/lib.rs
kklingenberg: This seems to be another pitfall I fell into. Consider the [man page](https://man7.org/linux/man-pages/man3/gamma.3.html):

> Because of historical variations in behavior across systems, this function is not specified in any recent standard. 4.2BSD had a gamma() that computed ln(|Gamma(|x|)|), leaving the sign of Gamma(|x|) in the external integer signgam.  In 4.3BSD the name was changed to [lgamma(3)](https://man7.org/linux/man-pages/man3/lgamma.3.html), and the man page promises
>
> "At some time in the future the name gamma will be rehabilitated and used for the Gamma function"
>
> This did indeed happen in 4.4BSD, where gamma() computes the Gamma function (with no effect on signgam).  However, this came too late, and we now have [tgamma(3)](https://man7.org/linux/man-pages/man3/tgamma.3.html), the "true gamma" function.

I tested this with the statically linked jq that can be downloaded from <https://jqlang.github.io/jq/>, and compared it to gojq. This is the result:

```text
$ jq -nc '1.5, 2.5, 3.5 | [tgamma, gamma, lgamma]'
[0.886226925452758,-0.12078223763524522,-0.12078223763524522]
[1.329340388179137,0.2846828704729192,0.2846828704729192]
[3.3233509704478426,1.2009736023470743,1.2009736023470743]
```

Vs.

```text
$ gojq -nc '1.5, 2.5, 3.5 | [tgamma, gamma, lgamma]'
[0.8862269254527579,0.8862269254527579,-0.12078223763524522]
[1.329340388179137,1.329340388179137,0.2846828704729192]
[3.323350970447843,3.323350970447843,1.2009736023470743]
```

So depending on how backwards-compatible we want to be, we should either make `gamma` a synonym of `lgamma` or `tgamma` 🥴

jaq-native/src/lib.rs
kklingenberg: I now see that (in the statically linked version of jq downloadable from the main site), the `scalb` function yields `nan` when the second argument is not an integer, while `scalbln` coerces the second argument to an integer.

```text
$ jq -n 'scalb(2.5; 2), scalbln(2.5; 2), scalbln(2.5; 2.99), (scalb(2.5; 2.99) | isnan)'
10
10
10
true
```

kklingenberg: I agree! And now everything clicked in my head 😆 the `ln` suffix stands for long! I kept assuming a logarithm was involved somewhere.

jaq-std/src/std.jq
kklingenberg: Done!

jaq-native/src/lib.rs
01mf02: In any case, we do not have to declare gamma here because we could just declare it in jaq-std. However, I propose for now that we simply do not declare gamma at all, until it is clarified what it should evaluate to.

jaq-native/tests/tests.rs
```diff
@@ -188,6 +188,53 @@ yields!(
     json!([[1], 2, 3])
 );
 
+#[test]
+fn math() {
```
01mf02, line 192: Could you split up this function into multiple invocations of the yields! macro?  That way, the tests are getting shorter, and you should be able to frequently omit the json! macro.

jaq-std/src/std.jq
```diff
@@ -22,6 +22,20 @@ def isinfinite: . == infinite or  . == -infinite;
...
+def logb:
+    if . == 0.0 then -infinite
+  elif isinfinite then infinite
+  elif isnan then .
+  else ilogb | . + 0.0 end;
+def pow10:       pow(10.0; .);
+def scalb(x; e): x * pow(2.0; e);
+def significand:
+    if isinfinite or isnan then .
+  elif . == 0.0 then 0.0
```
01mf02, line 35: You could merge this with the line above.

jaq-std/tests/std.rs
```diff
@@ -127,6 +127,37 @@ yields!(
     false
 );
 
+yields!(logb_nan, "nan | logb | isnan", true);
+yields!(logb_inf, "infinite | logb | isinfinite and . > 0", true);
+yields!(
+    logb_neg_inf,
+    "-infinite | logb | isinfinite and . > 0",
+    true
+);
```
01mf02, line 136: You could just compare with infinite and -infinite here.

jaq-std/tests/std.rs
```diff
@@ -127,6 +127,37 @@ yields!(
...
+yields!(
+    significand_neg_inf,
+    "-infinite | significand | isinfinite and . < 0",
+    true
+);
+yields!(
+    significand_range,
+    "[-123.456, -2.2, -2, -1, 0, 0.00001, 1, 2, 2.2, 123.456] | map(significand)",
+    [-1.929, -1.1, -1.0, -1.0, 0.0, 1.31072, 1.0, 1.0, 1.1, 1.929]
+);
```
01mf02, line 159: Great test!

jaq-std/src/std.jq
```diff
@@ -22,6 +22,20 @@ def isinfinite: . == infinite or  . == -infinite;
...
 def isnormal:   isnumber and ((. == 0 or isnan or isinfinite) | not);
 
+# Math
+def logb:
+    if . == 0.0 then -infinite
+  elif isinfinite then infinite
+  elif isnan then .
+  else ilogb | . + 0.0 end;
+def pow10:       pow(10.0; .);
+def scalb(x; e): x * pow(2.0; e);
```
01mf02, line 32: Could you add a test for this?

jaq-std/src/std.jq
```diff
@@ -22,6 +22,20 @@ def isinfinite: . == infinite or  . == -infinite;
...
+def significand:
+    if isinfinite or isnan then .
+  elif . == 0.0 then 0.0
+  else scalbln(.; ilogb | -1 * .) end;
+def drem($l; r): remainder($l; r) | if . == 0 then copysign(.; $l) else . end;
```
01mf02, line 37: Could you add a test for this?

jaq-native/src/lib.rs
```diff
@@ -260,6 +269,69 @@ fn now() -> Result<f64, Error> {
...
+    // drem is implemented in jaq-std
+    math::ff_f!(fdim),
+    math::ff_f!(fmax),
+    math::ff_f!(fmin),
+    math::ff_f!(fmod),
+    math::ff_f!(hypot),
+    math::if_f!(jn),
+    math::fi_f!(ldexp),
+    math::ff_f!(nextafter),
+    math::ff_f!("nexttoward", nextafter),
```
01mf02, line 326: This would also be better off in the standard library.

jaq-std/src/std.jq
kklingenberg: I wrote it like this to ensure the output type is always float :thinking:. It otherwise yields floats except when given `0` as an integer.

01mf02: Ah, I see. Thanks for clarifying this.

jaq-std/tests/std.rs
```diff
@@ -127,6 +127,60 @@ yields!(
...
+
+yields!(
+    scalb_nan,
+    "[scalb(nan, 1; nan, 1)] == [nan, nan, nan, 2.0]",
+    true
+);
+yields!(
+    scalb_range,
+    "[scalb(-2.5, 0, 2.5; 2, 3)]",
```
01mf02, line 158: It would be good to test this filter with float values as second arguments.

01mf02:
Could you update the README to include the new math filters?

jaq-std/tests/std.rs
kklingenberg: I agree, though in general I'm wary about including exact matches with approximations to irrational numbers. I guess I could include a test similar to what I did with `sin` and `cos` (which uses the pythagorean identity). Another option is to test for intervals, e.g. I know that `scalb(1; 1) < scalb(1; 1.5) < scalb(1; 2)`.

This is the reason for the next test: I can compare the results with an equivalent implementation using `pow`. Still, I'm not sure about either approach. If you're OK with having exact matches against fully-expanded f64s, then I'll include those cases as well.

kklingenberg:
> Could you update the README to include the new math filters?

Sure! Should I mention that all math filters excepting `gamma` are implemented? Or should I detail each filter in a collapsible list?

jaq-std/tests/std.rs
01mf02: How about something like `[scalb(-2.5, 0, 2.5; 2, 3) * 1000 | round]`?

01mf02:
> Sure! Should I mention that all math filters excepting `gamma` are implemented? Or should I detail each filter in a collapsible list?

A list of filters implemented should suffice. I would not even mention `gamma`.

jaq-std/tests/std.rs
kklingenberg: Brilliant!

01mf02:
We did it! Thanks for your long wind and all your effort.

kklingenberg:
🤝 Thank you for your support and guidance! I hope your shoulder heals well

PR #100, opened 2023-07-31
Try-catch

kklingenberg:
This PR adds the `try <f> catch <e>` syntax. It's still very much a work in progress.

- [x] Implement syntax
- [x] Add tests
- [x] Add benchmarks


jaq-core/src/error.rs
```diff
@@ -56,6 +56,15 @@ pub enum Error {
...
 
+impl Error {
+    /// Convert an error into a Val, to be used for `catch`
+    /// expressions.
+    pub fn as_val(&self) -> Val {
+        // TODO: make it better, Jude
+        Val::Null
+    }
+}
+
```
01mf02, line 67: You should not need this function; just use to_string().

jaq-core/src/filter.rs
```diff
@@ -204,6 +205,26 @@ impl<'a> FilterT<'a> for Ref<'a> {
...
+                        stream = if let Some(catch_filter) = catch_ {
+                            w(catch_filter).run((ctx.clone(), e.as_val()))
+                        } else {
+                            Box::new(std::iter::empty())
+                        };
+                        stream.next()
+                    }
+                    Some(y) => Some(y),
+                };
+                Box::new(std::iter::from_fn(step))
```
01mf02, line 226: You could drastically simplify your implementation by following the approach for the pipe operator `Ast::Pipe(l, false, r)`; that is, using `flat_map`.

01mf02:
Wow, I did not expect you working on this!
Just be careful with parsing try/catch, and be sure to test for nested try/catch expressions as well as what happens when you pass non-atomic expressions (e.g. f | g, 1 + 2) in try or catch.

jaq-core/src/error.rs
01mf02: "[The movement you need is on your shoulder](https://genius.com/983074/The-beatles-hey-jude/The-movement-you-need-is-on-your-shoulder)" -- damn those Beatles!

kklingenberg: What about the `Error::Val` case? I think we need this to work and yield `true`:` `try error(1) catch . + 1 == 2`. Anyway, I see that even then the `match` can simply be moved to the parser; I'll do that for now.

jaq-core/src/filter.rs
kklingenberg: I tried this, but failed in the following way: I need the error emitted inside the `try_` expression to short-circuit its stream. Flat-mapping doesn't allow that (meaning: I couldn't get that to work). That's why I wrote it the way I'd write it with the `stream!` macro and `yield`s: using the closest thing I could find which was `std::iter::from_fn`. I haven't pushed the `flat_map` version of this because of this test:

```text
$ jq -n 'try fma(2,3;1,1;1,"foo",2) catch .'
3
4
3
4
"string (\"foo\") number required"
```

Vs. my flat-mapping version:

```text
$ target/release/jaq.exe -n 'try fma(2,3;1,1;1,"foo",2) catch .'
3.0
"cannot use \"foo\" as float"
4.0
3.0
"cannot use \"foo\" as float"
4.0
4.0
"cannot use \"foo\" as float"
5.0
4.0
"cannot use \"foo\" as float"
5.0
```

jaq-core/src/filter.rs
```diff
@@ -37,6 +37,7 @@ pub enum Ast {
     Comma(Box<Self>, Box<Self>),
     Alt(Box<Self>, Box<Self>),
     Ite(Box<Self>, Box<Self>, Box<Self>),
+    TryCatch(Box<Self>, Option<Box<Self>>),
```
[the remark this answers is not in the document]
kklingenberg: Another instance of my lack of experience with large codebases :(

The `empty` filter is defined in `jaq-native`, which imports `jaq-core`. I need to provide `empty` in `lir.rs` in `jaq-core` (as far as I can tell). So that leaves me in a circular imports situation 😬 I can obviously simply provide the inline implementation of `empty`, which would be anti-DRY.

So humm, how would you proceed here? Should I make the `empty` filter an export of `jaq-core` somehow? That also sounds wrong

kklingenberg:
> [...] as well as what happens when you pass non-atomic expressions (e.g. f | g, 1 + 2) in try or catch.

What a can of worms this ended up being haha. Consider this:

```text
$ jq -n 'try "a" | error'
```
No error there. However:

```text
$ jq -n 'try "a" | error catch .'
jq: error: syntax error, unexpected catch, expecting $end
```

gojq is a bit more sane, and I'd consider it the canonical implementation on this:

```text
$ gojq -n 'try "a" | error'
gojq: error: a
```

And with the `catch`:

```text
$ gojq -n 'try "a" | error catch .'
gojq: invalid query: try "a" | error catch .
```

All of that weirdness goes away with parentheses. Still, the current state of this PR is incorrect in these regards. I'll try to give it another try tomorrow.

jaq-core/src/error.rs
01mf02: A potential test: `try (0|.[]) catch .`.

jaq-core/src/filter.rs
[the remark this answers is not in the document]
01mf02: Oh, I forgot that empty was not defined inside the core. My bad.
So we have two options: 
1. we could keep the Option the way you did
2. we could introduce `Ast::empty` the way it is already done with `Ast::recurse0`, simply yielding `[]|.[]`. That way, we could also remove Option from `Ast::Arr` to make construction of non-empty arrays faster, but making empty arrays slower.
What do you think?

jaq-core/src/filter.rs
01mf02: Another good catch.
How about putting a `take_while` before the `flat_map`?

01mf02:
> All of that weirdness goes away with parentheses.

I still consider it a historical error that try/catch is not terminated with `end` like if-then-else.

Anyway, good luck with parsing this in a sensible way. :)

01mf02:
I agree that the gojq approach is more sane. I would implement that.

jaq-core/src/filter.rs
kklingenberg: That's almost what I need. `take_while(Result::is_ok)` stops the first stream, but I need the non-ok value to feed into the `catch` filter. How do I do that? My google-fu is lacking.

My first attempt was including the "store the error in a mutable variable" part inside the closure for `take_while`, but then I hit the next wall: for the value to change I'd need to consume the first stream, which is a no-go considering streams might be infinite.

This does look like something that's already implemented in, say, Haskell's Data.List. Something like `take_while_then_chain_feeding_the_first_non_conformant_value(f)`.

jaq-core/src/filter.rs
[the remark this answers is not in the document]
kklingenberg: I'll take a stab at number 2. Thanks!

jaq-core/src/filter.rs
01mf02: Oh and there is another reason: d) I consider it an implicit design principle of jq to always use as many values as possible. That is, jq should not simply discard parts of a stream. To me, it feels that the semantics of `?`  and try/catch violate this principle.

01mf02: Furthermore, I would merge Try and TryCatch, because the former is a special case of the latter. For brevity, I would just call the whole thing `Try`.

kklingenberg: I'll follow your pointers. Still, just to try to justify jq's behaviour, I believe the short-circuiting is what enables this: <https://jqlang.github.io/jq/manual/#breaking-out-of-control-structures>. However, right after introducing the trick, the manual describes the keyword `break` which does the same thing natively.

01mf02: > However, right after introducing the trick, the manual describes the keyword `break` which does the same thing natively.

Yes, and I consider this explicit way of breaking the canonical way to do this example. I wouldn't be opposed, by the way, to implementing label/break, in contrast to implicit shortcutting in try/catch. But I believe that it will be a bit tricky.

jaq-core/src/filter.rs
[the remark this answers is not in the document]
01mf02: To make things easier for you, I implemented empty in be71aad.

kklingenberg:
I just applied the don't-break-out-of-streams criterion, and _tried_ (he he) to follow the gojq parser but I fear I ended up breaking everything in the parser. Tests pass though, so there's some hope.

Please do tell me if you'd rather take over from here. I initially thought this was low-hanging fruit, but now I'm starting to lose confidence. Also, the `label` / `break` feature looks frightening now, so I'll pass on that.

kklingenberg:
FYI, the current state of this PR performs quite poorly. These are the numbers, from a windows machine with an Intel i5 7400 (which is 6 years old):

| Benchmark | n | target/release/jaq | jq | gojq |
| ------------ | ------: | --------: | ---------: | ----------: |
| try-catch | 1048576 | 2.728 | 6.441 | 0.539 |

jaq-parse/src/filter.rs
```diff
@@ -46,6 +46,16 @@ where
...
+fn try_catch<P>(filter: P) -> impl Parser<Token, Spanned<Filter>, Error = P::Error> + Clone
+where
+    P: Parser<Token, Spanned<Filter>, Error = Simple<Token>> + Clone,
+{
+    let try_ = just(Token::Try).ignore_then(filter.clone().map(Box::new));
+    let catch_ = just(Token::Catch).ignore_then(filter.map(Box::new));
+    try_.then(catch_.or_not())
+        .map_with_span(|(try_, catch_), span| (Filter::TryCatch(try_, catch_), span))
+}
+
```
01mf02, line 58: Looks good!

jaq-parse/src/filter.rs
```diff
@@ -230,6 +242,7 @@ pub fn filter() -> impl Parser<Token, Spanned<Filter>, Error = Simple<Token>> +
         path.map_with_span(|(f, path), span| Filter::path(f, path, span)),
         fold(with_comma.clone()),
         if_then_else(with_comma.clone()),
+        try_catch(sans_comma_pipe.clone()),
```
01mf02, line 245: Couldn't you just use `path` instead of `sans_comma_pipe` here?
Consider the following example:

~~~
$ jq -n 'try (try 0["0"] + 1["1"]) catch .'
"Cannot index number with string \"1\""
~~~

Here you can see that jq does not capture the addition in the inner try block, but if I understand your implementation correctly, it would capture the addition.

jaq-core/src/filter.rs
```diff
@@ -178,7 +178,13 @@ impl<'a> FilterT<'a> for Ref<'a> {
                             .map(Val::obj)
                     }),
             ),
-            Ast::Try(f) => Box::new(w(f).run(cv).filter(|y| y.is_ok())),
+            Ast::Try(f, c) => Box::new(w(f).run((cv.0.clone(), cv.1)).flat_map(move |y| {
+                if let Err(e) = y {
+                    w(c).run((cv.0.clone(), e.as_val()))
+                } else {
+                    box_once(y)
+                }
```
01mf02, line 186: `map_or_else`?

01mf02:
> Please do tell me if you'd rather take over from here. I initially thought this was low-hanging fruit, but now I'm starting to lose confidence.

Don't worry, this is coming along fine.
I know that parsing this may not look easy, but perhaps with my feedback on your parsing code, we can make things work.

> Also, the `label` / `break` feature looks frightening now, so I'll pass on that.

Totally understandable. :) I would not venture into this myself until there is some significant demand for this feature.

README.md
```diff
@@ -641,6 +641,23 @@ it requires completely separate logic from `foreach/2` and `reduce`
...
+loops
+([here](https://jqlang.github.io/jq/manual/#breaking-out-of-control-structures)). jaq
+however doesn't interrupt the `f` stream, but instead sends _each_
+error value emitted to the `g` filter; the result is a stream of
+values emitted from `f` with values emitted from `g` interspersed
+where errors ocurred.
+
+```jq
+[try (1, error(2), 3, error(4)) catch .] == [1, 2, 3, 4]
+```
```
01mf02, line 658: I would clarify that this output is from jaq, and I would also include the output of jq for the same example and mark it accordingly.

01mf02:
> EDIT (linux testing): Tried it after compiling agains glibc on linux, and performance is somewhat different w.r.t. to jq and gojq:
> Benchmark 	n 	target/release/jaq 	jq 	gojq
> try-catch 	1048576 	0.332 	0.865 	0.436

To interpret these numbers, can you include the results of the other benchmarks as well (at least one that also generates a large sequence of numbers)?
Also, which version of jq did you use? I sometimes measured significant differences between jq 1.6 and the newest version on GitHub. 

> Weirdly, gojq reports _less real time than user time_ on this example, and beats jaq on real time as well (the numbers above are user time as per `bench.sh`). I didn't know that was possible, unless gojq uses parallel execution (which would be very cool). Anyway, for all of this to be meaningful I'd have to test the state of jaq before this PR as well. ~I'll try to do that soon.~ Which is impossible with the try-catch example since before this PR try-catch wasn't implemented.

Ah, benchmarking. The ultimate rabbit hole. :)
I wouldn't worry too much about this right now, because I don't think we can significantly speed this up anyway.

kklingenberg:
> can you include the results of the other benchmarks as well?

Will do once I get back to the linux machine, later tonight (or tomorrow, depending on timezone). Thank you for the inputs!

jaq-parse/src/filter.rs
kklingenberg: Indeed! I somehow thought that the only infix operators excluded were comma and pipe. Silly me, but also maybe I got distracted by the weird "pipe error" case in jq. Hopefully this unwinds all the damage I did in the parser 😅

01mf02:
I have thought a bit about this and I think that try/catch should actually capture any expression that is not a binary operation. This also includes the question mark operator and negation:

~~~
$ jq -n 'try 0[0]? catch .'
$ jq -n 'try -"a" catch .'
"string (\"a\") cannot be negated"
~~~

So I think that you could go with the following:

~~~
let neg = neg(try_).boxed();
let try_catch = try_catch(neg);
~~~

And then use try_catch instead of neg afterwards.
This way, you also shouldn't need to define a new recursive parser.

I do apologize by the way for the suboptimal naming of variables in the parser. This might have led to some confusion, for example `named`.

kklingenberg:
> So I think that you could go with the following:

This causes the following test to fail at the parsing phase:

```jq
try try {}[0] catch {}[1] catch .
```

This test has helped me _catch_ (he he) many a parsing mistake. Still, thanks for pointing out the prefix and postfix operation cases. I'll add them as tests.

01mf02:
Thanks again for all your efforts!
(I took the liberty to refine the parser a bit in https://github.com/01mf02/jaq/pull/100/commits/9a37978867cc23c25dbda593bad5402c71397ff9.)
Have a good weekend! :)

01mf02:
> This causes the following test to fail at the parsing phase:

You were right. Thank you for spotting this.

PR #102, opened 2023-08-05
ci: upload jaq executable to GitHub releases

eitsupi:
Add a new workflow to upload jaq executable files to GitHub releases.

I have tested on my fork.

image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,7 +50,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y gcc-aarch64-linux-gnu
           echo 'CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER=aarch64-linux-gnu-gcc' >>"$GITHUB_ENV"
-          echo 'RUSTFLAGS=-Zgcc-ld=lld' >>"$GITHUB_ENV"
+          echo 'CFLAGS=-mno-outline-atomics' >>"$GITHUB_ENV"
           echo 'CC=aarch64-linux-gnu-gcc' >>"$GITHUB_ENV"
       - name: Build
         run: cargo build --release --locked --target ${{ matrix.target }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,78 @@
+name: release
+on:
+  release:
+    types:
+      - released
+  workflow_dispatch:
+
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: build ${{ matrix.target }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+          - os: ubuntu-latest
+            target: aarch64-unknown-linux-musl
+          - os: macos-latest
+            target: x86_64-apple-darwin
+          - os: macos-latest
+            target: aarch64-apple-darwin
+          - os: windows-latest
+            target: x86_64-pc-windows-msvc
+    steps:
+      - uses: actions/checkout@v3
+      - run: rustup target add ${{ matrix.target }}
+        shell: bash
+      - if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y musl-tools
+      - if: matrix.target == 'aarch64-unknown-linux-musl'
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y gcc-aarch64-linux-gnu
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,85 @@
...
+        run: |
+          echo 'RUSTFLAGS=-Ctarget-feature=+crt-static' >>"$GITHUB_ENV"
+      - name: Build
+        if: runner.os != 'Linux'
+        run: cargo build --release --locked --target ${{ matrix.target }}
+        shell: bash
+      - name: Build with cross
+        if: runner.os == 'Linux'
+        run: cross build --release --locked --target ${{ matrix.target }}
+        shell: bash
```
01mf02, line 55: This looks redundant?

.github/workflows/release.yml
```diff
@@ -0,0 +1,85 @@
...
+          tar czf "../../../${ARTIFACT_NAME}" jaq ../../../LICENSE-MIT -C ../../../ README.md
+
+      - name: Create artifact for Windows
+        shell: bash
+        if: runner.os == 'Windows'
+        run: |
+          export ARTIFACT_NAME="jaq-${{ github.ref_type == 'tag' && github.ref_name || 0 }}-${{ matrix.target }}.zip"
+          echo "ARTIFACT_NAME=${ARTIFACT_NAME}" >>"$GITHUB_ENV"
+          cd target/${{ matrix.target }}/release
+          7z a "../../../${ARTIFACT_NAME}" jaq.exe ../../../LICENSE-MIT ../../../README.md
```
01mf02, line 72: Can you merge these two steps? Perhaps using something like:

~~~
        with:
          version: ${{ github.ref_type == 'tag' && github.ref_name || 0 }}
          archive: ${{ runner.os == 'Windows' && 'zip' || 'tar.gz' }}
          # ... (similar for 7z vs. tar, .exe etc.)
        run: |
          export ARTIFACT_NAME="jaq-${{ version }}-${{ matrix.target }}.${{ archive }}"
          # ...
~~~

.github/workflows/release.yml
```diff
@@ -0,0 +1,85 @@
...
+        run: cross build --release --locked --target ${{ matrix.target }}
+        shell: bash
+      - name: Create artifact for Linux and macOS
+        shell: bash
+        if: runner.os != 'Windows'
+        run: |
+          export ARTIFACT_NAME="jaq-${{ github.ref_type == 'tag' && github.ref_name || 0 }}-${{ matrix.target }}.tar.gz"
+          echo "ARTIFACT_NAME=${ARTIFACT_NAME}" >>"$GITHUB_ENV"
+          cd target/${{ matrix.target }}/release
+          tar czf "../../../${ARTIFACT_NAME}" jaq ../../../LICENSE-MIT -C ../../../ README.md
```
01mf02, line 63: Omit the `cd` and adapt the paths accordingly?

01mf02:
Do you know how I can test this workflow without creating a release?

.github/workflows/release.yml
eitsupi: The commands are different if closely looked at because they use CROSS only on Linux.
It is certainly possible to combine them into one step using if, but I am concerned that it would reduce readability.

.github/workflows/release.yml
eitsupi: I am sorry but I do not know how to do it.

eitsupi (commented):
> Do you know how I can test this workflow without creating a release?

Unfortunately I don't think there is a way to do that.
I am publishing the results of running this workflow in my fork for now so you can check it out.
Or, you should be able to copy this branch to a new repository and run it yourself.

.github/workflows/release.yml
eitsupi: Sorry, I couldn't figure out how to do the same thing without using cd.
Feel free to edit this PR to modify it.

01mf02:
Hi @eitsupi, thanks again for this PR. Given that this is a quite security-sensitive change, I decided to fully reimplement this functionality in order to understand how to use this safely. This is done in #107. I will prefer my PR to yours and hope that you understand. Still thank you for the inspiration. :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Install cross
+        if: runner.os == 'Linux'
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cross
+      - name: Set up for Windows static build
+        if: runner.os == 'Windows'
+        shell: bash
+        run: |
+          echo 'RUSTFLAGS=-Ctarget-feature=+crt-static' >>"$GITHUB_ENV"
```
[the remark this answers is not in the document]
01mf02: I just checked, and neither ripgrep nor hyperfine (both being fairly popular tools written in Rust) seem to include the flags you mentioned. I then went ahead and renamed `vcruntime140.dll` in `~/.wine/drive_c/windows/system32/` to something else, and the Windows binary I compiled with `cross` still executed fine. So I suppose that this is not an issue.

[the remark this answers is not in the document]
eitsupi: I see, thanks!

PR #105, opened 2023-08-09
Add formatters.

kklingenberg:
This PR adds the `@` filters that usually format things as strings using specific encodings, but also some that decode strings already encoded.

It is the absolute laziest version of this PR. I won't disagree if you decide to close this right away, or park it forever. It is lazy because:

- I simply added libraries for every encoding, and didn't attempt to implement any myself.
- The `@csv` and `@tsv` implementations are especially bad because they allocate csv::Writer for every element of the input stream.
- Extended the parser to allow for identifiers starting with `@`, because I didn't find an elegant way of limiting the `@` symbol to core-filters only. So users can define filters starting with `@` now, which is not possible in jq or gojq.
- ~I added a bunch of non-standard core filters that are referenced by the std implementations of the `@` filters.~
- ~I added a bunch of specific error variants, to capture Err variants from the various libraries that I wasn't quite sure I could just unwrap().~

If you still want to salvage this you can take over, or point me to ways of fixing stuff.

### DIFF
--- a/jaq-core/src/lib.rs
+++ b/jaq-core/src/lib.rs
@@ -142,6 +142,21 @@ where
     f(s, other).map_or_else(|| s.clone(), |stripped| Rc::new(stripped.into()))
 }
 
+fn fmt_row(v: &Val, f: impl Fn(&str) -> String) -> Result<String, Error> {
+    let fail = || alloc::format!("invalid value in a table row: {v}");
+    Ok(match v {
+        Val::Null => "".to_owned(),
+        Val::Str(s) => f(s),
+        Val::Arr(_) | Val::Obj(_) => return Err(Error::from_any(fail())),
+        v => v.to_string(),
+    })
+}
+
+fn to_csv(vs: &[Val]) -> Result<String, Error> {
+    let fr = |v| fmt_row(v, |s| alloc::format!("\"{}\"", s.replace('"', "\"\"")));
+    Ok(vs.iter().map(fr).collect::<Result<Vec<_>, _>>()?.join(","))
+}
+
 const CORE_RUN: &[(&str, usize, RunPtr)] = &[
     ("inputs", 0, |_, cv| {
         Box::new(cv.0.inputs().map(|r| r.map_err(Error::Parse)))
@@ -257,6 +272,13 @@ const CORE_RUN: &[(&str, usize, RunPtr)] = &[
             })))
         }))
     }),
+    ("@text", 0, |_, cv| {
+        box_once(Ok(Val::str(cv.1.to_string_or_clone())))
+    }),
+    ("@json", 0, |_, cv| box_once(Ok(Val::str(cv.1.to_string())))),
+    ("@csv", 0, |_, cv| {
+        box_once(cv.1.as_arr().and_then(|a| to_csv(a)).map(Val::str))
+    }),
 ];
 
 #[cfg(feature = "std")]
@@ -270,6 +292,66 @@ fn now() -> Result<f64, Error> {
 
 #[cfg(feature = "std")]
 const STD: &[(&str, usize, RunPtr)] = &[("now", 0, |_, _| box_once(now().map(Val::Float)))];
+
+#[cfg(feature = "format")]
+fn to_tsv(vs: &[Val]) -> Result<String, Error> {
+    let fs = |s: &str| {
+        let patterns = &["\n", "\r", "\t", "\\"];
+        let replace_with = &["\\n", "\\r", "\\t", "\\\\"];
+        let ac = aho_corasick::AhoCorasick::new(patterns).unwrap();
+        ac.replace_all(s, replace_with)
+    };
+    let fr = |v| fmt_row(v, fs);
+    Ok(vs.iter().map(fr).collect::<Result<Vec<_>, _>>()?.join("\t"))
+}
+
+#[cfg(feature = "format")]
+const FORMAT: &[(&str, usize, RunPtr)] = &[
+    ("@tsv", 0, |_, cv| {
+        box_once(cv.1.as_arr().and_then(|a| to_tsv(a)).map(Val::str))
+    }),
+    ("@html", 0, |_, cv| {
+        box_once(Ok(Val::str(
+            html_escape::encode_safe(&cv.1.to_string_or_clone()).into_owned(),
+        )))
+    }),
+    ("@uri", 0, |_, cv| {
+        box_once(Ok(Val::str(
+            urlencoding::encode(&cv.1.to_string_or_clone()).into_owned(),
+        )))
+    }),
+    ("@urid", 0, |_, cv| {
+        box_once(
+            urlencoding::decode(&cv.1.to_string_or_clone())
+                .map_err(Error::from_any)
+                .map(|s| Val::str(s.into_owned())),
+        )
+    }),
+    ("@sh", 0, |_, cv| {
+        box_once(Ok(Val::str(
+            shell_escape::escape(cv.1.to_string_or_clone().into()).into_owned(),
+        )))
+    }),
+    ("@base64", 0, |_, cv| {
+        use base64::{engine::general_purpose, Engine as _};
+        box_once(Ok(Val::str(
+            general_purpose::STANDARD.encode(cv.1.to_string_or_clone()),
+        )))
+    }),
+    ("@base64d", 0, |_, cv| {
+        use base64::{engine::general_purpose, Engine as _};
+        box_once(
+            general_purpose::STANDARD
+                .decode(cv.1.to_string_or_clone())
+                .map_err(Error::from_any)
+                .and_then(|d| {
+                    std::str::from_utf8(&d)
+                        .map_err(Error::from_any)
+                        .map(|s| Val::str(s.to_owned()))
+                }),
+        )
+    }),
+];
 
 #[cfg(feature = "math")]
 const MATH: &[(&str, usize, RunPtr)] = &[
@@ -404,86 +486,3 @@ const LOG: &[(&str, usize, RunPtr, UpdatePtr)] = &[(
     |_, cv| box_once(Ok(debug(cv.1))),
     |_, cv, f| f(debug(cv.1)),
 )];
-
-fn to_csv(vs: &[Val]) -> Result<String, Error> {
-    let fail = |v: &Val| alloc::format!("invalid value in a CSV row: {v}");
-    let f = |v: &Val| match v {
-        Val::Null => Ok("".to_owned()),
-        Val::Str(s) => Ok(alloc::format!("\"{}\"", s.replace('"', "\"\""))),
-        Val::Arr(_) | Val::Obj(_) => Err(Error::from_any(fail(v))),
-        v => Ok(v.to_string()),
-    };
-    Ok(vs.iter().map(f).collect::<Result<Vec<_>, _>>()?.join(","))
-}
-
-#[cfg(feature = "format")]
-fn to_tsv(vs: &[Val]) -> Result<String, Error> {
-    let fail = |v: &Val| alloc::format!("invalid value in a TSV row: {v}");
-    let f = |v: &Val| match v {
-        Val::Null => Ok("".to_owned()),
-        Val::Str(s) => {
-            let patterns = &["\n", "\r", "\t", "\\"];
-            let replace_with = &["\\n", "\\r", "\\t", "\\\\"];
-            let ac = aho_corasick::AhoCorasick::new(patterns).unwrap();
-            Ok(ac.replace_all(s, replace_with))
-        }
-        Val::Arr(_) | Val::Obj(_) => Err(Error::from_any(fail(v))),
-        v => Ok(v.to_string()),
-    };
-    Ok(vs.iter().map(f).collect::<Result<Vec<_>, _>>()?.join("\t"))
-}
-
-#[cfg(feature = "format")]
-const FORMAT: &[(&str, usize, RunPtr)] = &[
-    ("@text", 0, |_, cv| {
-        box_once(Ok(Val::str(cv.1.to_string_or_clone())))
-    }),
-    ("@json", 0, |_, cv| box_once(Ok(Val::str(cv.1.to_string())))),
-    ("@csv", 0, |_, cv| {
-        box_once(cv.1.as_arr().and_then(|a| to_csv(a)).map(Val::str))
-    }),
-    ("@tsv", 0, |_, cv| {
-        box_once(cv.1.as_arr().and_then(|a| to_tsv(a)).map(Val::str))
-    }),
-    ("@html", 0, |_, cv| {
-        box_once(Ok(Val::str(
-            html_escape::encode_safe(&cv.1.to_string_or_clone()).into_owned(),
-        )))
-    }),
-    ("@uri", 0, |_, cv| {
-        box_once(Ok(Val::str(
-            urlencoding::encode(&cv.1.to_string_or_clone()).into_owned(),
-        )))
-    }),
-    ("@urid", 0, |_, cv| {
-        box_once(
-            urlencoding::decode(&cv.1.to_string_or_clone())
-                .map_err(Error::from_any)
-                .map(|s| Val::str(s.into_owned())),
-        )
-    }),
-    ("@sh", 0, |_, cv| {
-        box_once(Ok(Val::str(
-            shell_escape::escape(cv.1.to_string_or_clone().into()).into_owned(),
-        )))
-    }),
-    ("@base64", 0, |_, cv| {
-        use base64::{engine::general_purpose, Engine as _};
-        box_once(Ok(Val::str(
-            general_purpose::STANDARD.encode(cv.1.to_string_or_clone()),
-        )))
-    }),
-    ("@base64d", 0, |_, cv| {
-        use base64::{engine::general_purpose, Engine as _};
-        box_once(
-            general_purpose::STANDARD
-                .decode(cv.1.to_string_or_clone())
-                .map_err(Error::from_any)
-                .and_then(|d| {
-                    std::str::from_utf8(&d)
-                        .map_err(Error::from_any)
-                        .map(|s| Val::str(s.to_owned()))
-                }),
-        )
-    }),
-];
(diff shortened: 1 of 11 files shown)


jaq-core/src/lib.rs
```diff
@@ -402,3 +404,93 @@ const LOG: &[(&str, usize, RunPtr, UpdatePtr)] = &[(
     |_, cv| box_once(Ok(debug(cv.1))),
     |_, cv, f| f(debug(cv.1)),
 )];
+
+#[cfg(feature = "format")]
+fn to_csv(v: Val, delimiter: u8) -> ValR {
```
01mf02, line 409: Here, it would be preferable to accept a `&[Val]` and to call this function on `v.as_str()`.

jaq-core/src/lib.rs
```diff
@@ -402,3 +404,93 @@ const LOG: &[(&str, usize, RunPtr, UpdatePtr)] = &[(
...
+                .from_writer(vec![]);
+            let mut record = StringRecord::new();
+            for cell in &*vs {
+                match cell {
+                    Val::Str(s) => {
+                        record.push_field(s);
+                        Ok(())
+                    }
+                    _ => Err(Error::Str(cell.clone())),
+                }?;
```
01mf02, line 424: You could shorten this a lot by using `cell.as_str()`.

jaq-core/src/lib.rs
```diff
@@ -402,3 +404,93 @@ const LOG: &[(&str, usize, RunPtr, UpdatePtr)] = &[(
...
+        box_once(match cv.1 {
+            Val::Str(s) => Ok(Val::str(html_escape::encode_safe(&*s).to_string())),
+            _ => Err(Error::Str(cv.1)),
+        })
+    }),
+    ("touri", 0, |_, cv| {
+        box_once(match cv.1 {
+            Val::Str(s) => Ok(Val::str(urlencoding::encode(&s).to_string())),
+            _ => Err(Error::Str(cv.1)),
+        })
```
01mf02, line 453: You frequently have this pattern where you check if the input value is a string, and if it is, then run some function on the string and return another string.
I would create a function in `Val` that encodes this pattern, similarly to `mutate_str()`. You could call it `map_str`, for example. For functions that can fail during this conversion, like `fromuri`, you could introduce another function like `try_map_str()`.

01mf02:
Thanks a lot, this is pretty cool work!
It is quite funny because I was recently also pondering doing something very similar to what you have done now. In particular, like you, I came to the conclusion that it is the easiest way to just allow defining filters whose name starts with an `@`. So no objections from me on this front.
I also think that it is good style to salvage existing crates that perform these conversions instead of writing that stuff ourselves. That way, there are less things to get wrong.
My main objection would be your declaration of new error variants. I have started to rework the Error type, by removing all error variants that are not produced in jaq-interpret itself. Until I have merged this new error type, could you encode errors by writing them into `Error::Val(Val::str(..))`?

jaq-core/src/lib.rs
01mf02: This remark of mine is probably obsolete if you consider [my comment below](https://github.com/01mf02/jaq/pull/105#discussion_r1289774934).

jaq-core/src/lib.rs
01mf02: This remark of mine is probably obsolete if you consider [my comment below](https://github.com/01mf02/jaq/pull/105#discussion_r1289774934).

kklingenberg:
I think I managed to follow your pointers. However clippy was complaining about naming `Error::from_str` that way so I just used the name `Error::from_any`. Which is obscure, and _worse_, but now clippy doesn't complain.

jaq-std/src/std.jq
```diff
@@ -147,3 +147,15 @@ def input: first(inputs);
...
+def    @text: tostring;
+def    @json: tojson;
+def    @html: tostring | tohtml;
+def     @uri: tostring | touri;
+def    @urid: tostring | fromuri;
+def     @csv: map(tostring) | tocsv;
+def     @tsv: map(tostring) | totsv;
+def      @sh: tostring | tosh;
+def  @base64: tostring | tobase64;
+def @base64d: tostring | frombase64;
```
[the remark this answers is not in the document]
kklingenberg: Done!

jaq-core/src/lib.rs
kklingenberg: Done!

jaq-core/src/lib.rs
```diff
@@ -402,3 +404,82 @@ const LOG: &[(&str, usize, RunPtr, UpdatePtr)] = &[(
     |_, cv| box_once(Ok(debug(cv.1))),
     |_, cv, f| f(debug(cv.1)),
 )];
+
+#[cfg(feature = "format")]
+fn stringify(v: Val) -> String {
+    v.clone()
+        .to_str()
```
01mf02, line 411: It's better to use as_str() here and clone only conditionally.

jaq-core/src/lib.rs
```diff
@@ -402,3 +404,82 @@ const LOG: &[(&str, usize, RunPtr, UpdatePtr)] = &[(
...
+        .map_or_else(|_| v.to_string(), |s| s.to_string())
```
01mf02, line 412: I would replace s.to_string() by  something like s.clone()  here, so it is clear that we are really just reusing the string that is there and not creating an escaped version of it.

pkoppstein:
My impression (perhaps mistaken) is that this PR strays quite widely from jq's "@foo" operators.

As the jq manual says, "The @foo syntax is used to format and escape strings".  

This means that one can write expressions such as:
```
 "@uri "https://www.google.com/search?q=\(.search)"
```
That is, `@foo` can both be used as a 0-arity filter, or as a kind of "prefix operator".

jaq-core/src/lib.rs
```diff
@@ -402,3 +404,82 @@ const LOG: &[(&str, usize, RunPtr, UpdatePtr)] = &[(
     |_, cv| box_once(Ok(debug(cv.1))),
     |_, cv, f| f(debug(cv.1)),
 )];
+
+#[cfg(feature = "format")]
+fn stringify(v: Val) -> String {
```
01mf02, line 409: I would actually make this a member function of Val, because it is a functionality which will be used implicitly in the interpreter. That is, `"world" | "Hello, \(.)"` will use this function for interpolation.
However I would rename this function; what do you think about to_string_or_clone()? I know it's a bit longer than stringify, but at least it's a bit more descriptive and clarifies the difference with respect to to_string().

jaq-core/src/lib.rs
```diff
@@ -402,3 +404,82 @@ const LOG: &[(&str, usize, RunPtr, UpdatePtr)] = &[(
...
+    }),
+    ("@csv", 0, |_, cv| {
+        box_once(cv.1.as_arr().and_then(|vs| to_csv(vs, b',')))
+    }),
+    ("@tsv", 0, |_, cv| {
+        box_once(cv.1.as_arr().and_then(|vs| to_csv(vs, b'\t')))
+    }),
+    ("@sh", 0, |_, cv| {
+        box_once(Ok(Val::str(
+            shell_escape::escape(stringify(cv.1).into()).to_string(),
```
01mf02, line 463: Use into_owned() instead of to_string()?

jaq-core/src/lib.rs
```diff
@@ -402,3 +404,82 @@ const LOG: &[(&str, usize, RunPtr, UpdatePtr)] = &[(
...
+    }),
+    ("@urid", 0, |_, cv| {
+        box_once(
+            urlencoding::decode(&stringify(cv.1))
+                .map_err(Error::from_any)
+                .map(|s| Val::str(s.to_string())),
+        )
+    }),
+    ("@csv", 0, |_, cv| {
+        box_once(cv.1.as_arr().and_then(|vs| to_csv(vs, b',')))
```
01mf02, line 456: Can you add a test that shows what happens when the input contains commas (inside a string)?

jaq-core/src/lib.rs
```diff
@@ -402,3 +404,82 @@ const LOG: &[(&str, usize, RunPtr, UpdatePtr)] = &[(
...
+    ("@base64d", 0, |_, cv| {
+        use base64::{engine::general_purpose, Engine as _};
+        box_once(
+            general_purpose::STANDARD
+                .decode(stringify(cv.1))
+                .map_err(Error::from_any)
+                .and_then(|d| {
+                    std::str::from_utf8(&d)
+                        .map_err(Error::from_any)
+                        .map(|s| Val::str(s.to_string()))
```
01mf02, line 481: Use map_or_else() and to_ owned()?

jaq-parse/src/token.rs
```diff
@@ -106,6 +106,8 @@ pub fn token() -> impl Parser<char, Token, Error = Simple<char>> {
 
     let var = just('$').ignore_then(text::ident());
 
+    let format_filter = just('@').chain(text::ident()).collect();
```
01mf02, line 109: Could you use `just('@').or_not()` to make this more general and then use it instead of text::ident() to define ident below?

01mf02:
> That is, `@foo` can both be used as a 0-arity filter, or as a kind of "prefix operator".

Correct. This PR adds support for the first case only. However, I do already have an idea how this can be extended in the future in order to also support the second case, that is, string interpolation with custom formatting as prefix operator like in your example.
I suppose that you will be happy to hear this, given that you have been missing string interpolation so badly. :) It's not that far away anymore. However, I want to release 1.0 without string interpolation yet, given how many new features made it into this version already.

pkoppstein:
@01mf02 - Thanks for your fast response, which was very welcome in several respects!

Perhaps you missed my implied point, which was to emphasize that identifiers starting with "@" are (in jq) very special in that they can be used as prefix operators.  Of course, whatever potential confusion this might introduce in jaq can be mitigated by suitable documentation.

jaq-core/src/lib.rs
kklingenberg: As a testament to my inexperience with Rust library design, I really screwed this up w.r.t. using `Rc` and cloning everywhere. I tried the `Cow` route and the `AsRef` route, but failed to make it "nice" when calling the function and `Val::str`.

Anyway, feel free to rework it all to reduce unnecessary cloning and such.

01mf02:
> Perhaps you missed my implied point, which was to highlight the tension that is created by the decision to allow "users [to] define filters starting with @".
> 
> That is, identifiers starting with "@" are (in jq) very special in that they can be used as prefix operators.

I plan to keep this particular behaviour; that is, only filters whose name starts with an @ should be usable as string prefix operators.
Still, @-identifiers will be more flexible in jaq. For example, I plan to allow filters starting with an @ that take arguments, also when used in prefix form. I do not expect that this will be used a lot, but it would actually complicate the implementation more to prohibit it, so I don't see a good reason to disallow it. 
 
> Of course, whatever potential confusion this might introduce in jaq can be mitigated by suitable documentation.

I agree.

jaq-core/src/lib.rs
```diff
@@ -402,3 +404,76 @@ const LOG: &[(&str, usize, RunPtr, UpdatePtr)] = &[(
     |_, cv| box_once(Ok(debug(cv.1))),
     |_, cv, f| f(debug(cv.1)),
 )];
+
+#[cfg(feature = "format")]
+fn to_csv(vs: &[Val], delimiter: u8) -> ValR {
```
01mf02, line 409: Can you make this function return a string result instead of a value result? In general, when a function returns only one kind of value, I return this directly instead of wrapping it in a value.

jaq-interpret/src/path.rs
```diff
@@ -59,7 +59,9 @@ impl Part<Vec<Val>> {
                     let until = rel_bounds(until).map(move |i| Ok(abs_bound(i?, len, len)));
                     Box::new(prod(from, until).map(move |(from, until)| {
                         let (skip, take) = skip_take(from?, until?);
-                        Ok(Val::str(s.chars().skip(skip).take(take).collect()))
+                        Ok(Val::str(
+                            s.chars().skip(skip).take(take).collect::<String>(),
+                        ))
```
01mf02, line 64: This doesn't seem to be related to formatters?

01mf02: Ah, that's because you modified the str function.
I think that accepting parameters implementing Into should be done quite sparingly, as it can require more type annotations, like here. May I ask you for your motivation for having changed the signature of the function?

jaq-interpret/src/val.rs
```diff
@@ -55,8 +55,8 @@ fn rc_unwrap_or_clone<T: Clone>(a: Rc<T>) -> T {
 
 impl Val {
     /// Construct a string value.
-    pub fn str(s: String) -> Self {
-        Self::Str(s.into())
+    pub fn str(s: impl Into<String>) -> Self {
+        Self::Str(Rc::new(s.into()))
```
01mf02, line 59: I'm sorry, but I do not like this change because it makes the string function different from all other similar functions like arr, obj and so on. This might be useful, but only if all functions are changed accordingly, which I would however not do as part of this PR. Let's keep it small.

jaq-interpret/src/val.rs
```diff
@@ -109,6 +109,15 @@ impl Val {
         }
     }
 
+    /// If the value is a Str, extract the inner string, else convert
+    /// it to string.
+    pub fn to_string_or_clone(self) -> String {
+        match self {
+            Self::Str(s) => s.to_string(),
```
01mf02, line 116: Here I would use to_owned().
Otherwise, this will make people (like me ^^) reading this wonder whether this will escape the string.

jaq-parse/src/token.rs
```diff
@@ -110,23 +110,27 @@ pub fn token() -> impl Parser<char, Token, Error = Simple<char>> {
...
-        "for" => Token::For,
-        "foreach" => Token::Foreach,
-        "try" => Token::Try,
-        "catch" => Token::Catch,
-        _ => Token::Ident(ident),
-    });
+    let ident = just('@')
+        .or_not()
+        .chain::<char, String, _>(text::ident())
+        .collect()
```
01mf02, line 116: To reduce the size of this change, could you terminate the definition of ident after collect(), then redefine ident using the previously defined variable ident?

jaq-core/src/lib.rs
01mf02: Thanks!

kklingenberg:
So, about `map_or_else`: do you prefer it to `map_err().map()` even when the closures in `map_or_else` would produce `Err` and `Ok` values respectively? It somehow seems wrong to me, but I'd rather keep it consistent with your preferences.

jaq-core/src/lib.rs
01mf02: I don't see a lot of unnecessary cloning. Don't worry, I think you're doing a good job.

jaq-interpret/src/val.rs
kklingenberg: I'll have to go with `(*s).clone()` since `to_owned()` over an `Rc` yields another `Rc`. I think.

01mf02: That sounds good to me!

kklingenberg:
Amazingly, it took me this long to check de official definition of the `@sh` filter:

> @sh:
> The input is escaped suitable for use in a command-line for a POSIX shell. If the input is an array, the output will be a series of space-separated strings.

Also there's this PR in jq: <https://github.com/jqlang/jq/pull/2828>

So possibly that one requires a heavy rewrite. Perhaps it's best to wait for jq's PR to land?

01mf02:
@kklingenberg, I have a few comments regarding this PR:

*  I made a few formatters available without the “format” feature gate. This is because I use feature gates in jaq exclusively to reduce the number of external dependencies.
* I rewrote a few tests to use strings of the format r#”...”#. This is really important to respect because it simplifies testing the tests with jq by copy-pasting them into the command line. Tests that are hard to test are less likely to be tested.
* Speaking of the tests, if the behavior of jaq diverges from that of jq, this has to be documented. In particular, the HTML, CSV, and TSV behavior diverged in several ways, which I have now tried to remedy. Please make sure to compare your remaining implementations with jq; see `@sh`, but also `@uri` could be tested with more input.
* Speaking of @sh, I would implement the behavior in jq 1.6, that is, special casing for arrays and rejection of objects.
* You added a new filter @urid; this should probably be documented.

kklingenberg:
I thank you for all the effort you put into fixing this; I expected it would be messy from the start. It probably ended up being more work to guide me through this, than to simply implement it yourself.

I tried to emulate the behaviour of jq's `@sh`, but the library I used diverges from jq in that it doesn't escape what it doesn't need to, and escapes some symbols more aggresively (e.g. `!`). I also removed the non-standard `@urid` filter, which I had Initially pulled from gojq's HEAD.

Also about `@sh`, another thing that worries me is that the library has different behaviour for unix or windows: <https://docs.rs/shell-escape/latest/src/shell_escape/lib.rs.html#17-23>. I believe this may cause further divergence from jq.

Anyway, I hereby yield control of this PR, as I don't have the required motivation to continue. Thanks again!

01mf02:
> I thank you for all the effort you put into fixing this; I expected it would be messy from the start. It probably ended up being more work to guide me through this, than to simply implement it yourself.

Thank you for having attacked this functionality. Even if it took some of my time indeed, it is important for me that other people than me become familiar with my philosophy of working on jaq, and I hope that you learned something from working on this.

> I tried to emulate the behaviour of jq's `@sh`, but the library I used diverges from jq in that it doesn't escape what it doesn't need to, and escapes some symbols more aggresively (e.g. `!`).
> Also about `@sh`, another thing that worries me is that the library has different behaviour for unix or windows: https://docs.rs/shell-escape/latest/src/shell_escape/lib.rs.html#17-23. I believe this may cause further divergence from jq.

Platform-specific output is a no-go. It is also not what jq does, quoth the jq documentation: "The input is escaped suitable for use in a command-line *for a POSIX shell*."
In the end, I found that jq only escapes `'` characters, which was sufficiently easy to do without any supporting library.

> I also removed the non-standard `@urid` filter, which I had Initially pulled from gojq's HEAD.

That's good.

> Anyway, I hereby yield control of this PR, as I don't have the required motivation to continue. Thanks again!

Thanks to you too!

Issue #120, opened 2023-10-31
Failed to compile 1.1.0


mo8it:
I get this error when trying to update the binary with `cargo-update`:

```
   Compiling jaq v1.1.0
error[E0282]: type annotations needed
   --> /var/home/mo8it/.cargo/registry/src/index.crates.io-6f17d22bba15001f/jaq-1.1.0/src/main.rs:441:57
    |
441 |                 ColoredFormatter::new(CompactFormatter).write_colored_json(
    |                                                         ^^^^^^^^^^^^^^^^^^ cannot infer type of the type parameter `T` declared on the method `write_colored_json`
    |
help: consider specifying the generic arguments
    |
441 |                 ColoredFormatter::new(CompactFormatter).write_colored_json::<_, T>(
    |                                                                           ++++++++

For more information about this error, try `rustc --explain E0282`.
error: could not compile `jaq` (bin "jaq") due to previous error
error: failed to compile `jaq v1.1.0`, intermediate artifacts can be found at `/tmp/cargo-installnQRe5x`.
```

kklingenberg:
Could you include the cargo and rustc versions you're using?

```bash
cargo --version
rustc --version
```

mo8it:
I am using the latest stable Rust version on Linux.

`cargo --version`:

```
cargo 1.73.0
```

`rustc --version`:

```
rustc 1.73.0
```

The problem is probably because `cargo-update` doesn't use `--locked` by default. I know that the official installation instructions for `jaq` recommend using `--locked`, but many prefer installation without it to receive bug fixes.

I think that the problem will arise anyway after updating the dependencies.
Therefore, I would suggest updating the dependencies, fixing that small compiler error and publishing `1.1.2` ❤️

ctron:
I re-released the version 3.3.0 as 4.0.0 and also yanked the 3.3.0 release. Sorry for the troubles.

mo8it:
I can confirm that compilation works now after the version 3.3.0 is yanked. @ctron Thanks 🥰

PR #122, opened 2023-10-31
Update deps

mo8it:
Fixes https://github.com/01mf02/jaq/issues/120

The problem was the following commit that changed the type of the argument `value` from `Value` to a generic type that implements `Serialize`: https://github.com/ctron/colored_json/commit/df5c79b33e199a8bca7c02a1e5f29867604fe1bd

You can directly merge the following PR instead of merging this first if you want to upgrade all dependencies: https://github.com/01mf02/jaq/pull/123

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,9 +17,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.2"
+version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43f6cb1bf222025340178f382c426f13757b2960e89779dfcb319c32542a5a41"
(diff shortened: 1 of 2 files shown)


mo8it:
There is no need for this PR anymore since the problematic version of `colored_json` is yanked.

01mf02:
Thanks for this PR. I permitted myself to use parts of your commit aa71c3b2dda92a726659006bf7b121dd4cbef954 in ea4517aab468f7cf29aa568013aebf84be434d1a to make this future-proof, but I do not want to upgrade `colored_json`, because it would increase the MSRV.

mo8it:
There is no rush in merging https://github.com/01mf02/jaq/pull/123. Just merge whenever you want to raise the MSRV.

PR #123, opened 2023-10-31
Upgrade deps

mo8it:
Can be merged instead or in addition to https://github.com/01mf02/jaq/pull/122

It doesn't only run `cargo update`, but it also upgrades dependencies with breaking changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,9 +17,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.2"
+version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


01mf02:
Thanks for your effort! Sparked by your upgrade of `colored_json`, I noticed that jaq did not build anymore on Rust 1.62, which was the previous MSRV. That's why I stepped back and actually downgraded many dependencies, in order to achieve a MSRV of at least 1.65. This is now checked by a GitHub action.
So I'm sorry that I cannot accept your PR (because many updated packages do not build anymore with Rust 1.65), but I still would like to thank you for having taking the time!

mo8it:
Don't worry, I didn't spend a lot of time on this PR :P

I didn't ask before creating it anyway, so I created it knowing that it will possibly get closed.

PR #125, opened 2023-11-01
Fix typos

mo8it:
Easy fixes by running `typos -w`.

[`typos`](https://github.com/crate-ci/typos) is an awesome tool written in Rust :D


01mf02:
Good catches! Regarding [focusses](https://grammarhow.com/focusses-or-focuses/), that one is not technically wrong, but apparently getting rare.

PR #130, opened 2023-11-21
Recursive object merge

kklingenberg:
This PR implements `*` for objects, according to the spec here: <https://jqlang.github.io/jq/manual/#multiplication-division-modulo>


01mf02:
That looks very nice; thanks a lot!

01mf02:
I improved your implementation in f24da821c7d0ce286d3fe8a0c87fee5c9ed75268 by removing some `clone()` and by making it clear that recursive object merging never fails.

PR #132, opened 2023-11-29
Added support for WASIX and published it to Wasmer!

syrusakbary:
First of all, congrats on the awesome work... love what you are doing!

I compiled the project to [WASIX](https://wasix.org) to allow running `jaq` anywhere via WebAssembly (even on the browser!). I also published under [jaq/jaq](https://wasmer.io/jaq/jaq) in [Wasmer](https://wasmer.io) (please let me know if you have a Wasmer username so I can add you to the namespace).

This PR is fairly simple, but it allows running jaq just with `wasmer`:

```bash
echo '{"a": 1, "b": 2}' | wasmer run jaq/jaq '.a'
```

I'm happy to add CI integration as well, so it's automatically published in the future. Just let me know!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,4 +11,5 @@ resolver = "2"
 
 [profile.release]
 strip = true
-codegen-units = 1
+opt-level = 'z'
+lto = true
--- a/README.md
+++ b/README.md
@@ -77,6 +77,12 @@ jaq should work on any system supported by Rust.
 If it does not, please file an issue.
 
 
+You can also use [Wasmer](https://wasmer.io) to run `jaq` without an
+installation required, using the [jaq/jaq package](https://wasmer.io/jaq/jaq):
+
+    $ echo '{"a": 1, "b": 2}' | wasmer run jaq/jaq
+
+
 ## Binaries
 
 You may also install jaq using [homebrew](https://formulae.brew.sh/formula/jaq) on macOS or Linux:
--- a/README.md
+++ b/README.md
@@ -80,7 +80,9 @@ If it does not, please file an issue.
 You can also use [Wasmer](https://wasmer.io) to run `jaq` without an
 installation required, using the [jaq/jaq package](https://wasmer.io/jaq/jaq):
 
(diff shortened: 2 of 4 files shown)


01mf02:
Hi @syrusakbary, thanks for this interesting work!

I'm fine with most changes, but what is your motivation for your changes to `Cargo.toml`, in particular leaving out `codegen-units` and adding `lto`?
Also, how about making the version number of the Wasmer release match the actual jaq version number?

syrusakbary:
> How can I use it in browser?

Yes, thanks to the new released Wasmer SDK you can simply use the `jaq/jaq` package.

> I'm fine with most changes, but what is your motivation for your changes to Cargo.toml, in particular leaving out codegen-units and adding lto?

The main reason is that otherwise the Wasm file generated was not runnable by the runtime (it was generating a wasm file with functions too big to be processed by the Wasm runtime).

> Also, how about making the version number of the Wasmer release match the actual jaq version number?

That makes complete sense. I'll fix that!

Also, @01mf02 do you have an account in Wasmer so I can add you as admin to the package as well?

01mf02:
> The main reason is that otherwise the Wasm file generated was not runnable by the runtime (it was generating a wasm file with functions too big to be processed by the Wasm runtime).

I thought something like this. Is there a way to specify these flags so that they are only used for the WASM build?
(Otherwise they might have negative performance implications on the regular builds.)

> That makes complete sense. I'll fix that!

Great!

> Also, @01mf02 do you have an account in Wasmer so I can add you as admin to the package as well?

I just created one. It's called 01mf02. :)

01mf02:
@syrusakbary, I might have found a partial way towards running WASM builds with different flags. On [this article](https://blog.orhun.dev/stdout-vs-stderr/), I found the following syntax:

~~~
[profile.profiling]
inherits = "release"
debug = true
~~~

So we could make a profile "wasm" or something like that, let it inherit "release" and then set whatever other flags it needs.

01mf02:
There has unfortunately not been any progress on this PR for more than a year, so I'm closing this. If you are interested in working on the points discussed, feel free to ping me.

PR #137, opened 2023-11-30
Fixed typo

hmottestad:
Fixed a typo.


pkoppstein (changes requested):
Unfortunately, as a matter of English, the proposed correction is not quite right,
and certainly worse than the original.

An alternative and perhaps preferable way of expressing the original thought would be along the lines of:

This is particularly evident when ...

PR #138, opened 2023-11-30
Fix path parsing for expressions like `.a.[]`

### LINKED ISSUE #134: Unexpected token `[` in '.nodes.[]'
I have simplified query `.nodes.[]` which results in the following error:
```
> jaq '.nodes.[]'
Error: Unexpected token while parsing string, expected "
   ╭─[<unknown>:1:8]
   │
 1 │ .nodes.[]
   │        ┬  
   │        ╰── Unexpected token [
───╯
```

`jq` handles the query fine.

kklingenberg:
Should close #134


01mf02:
Hi @kklingenberg, thanks for your effort. However, as I explain in my own PR #145, I wanted to take this opportunity to completely overdo the path parsing. Still thanks for your time and your tests!

Issue #152, opened 2024-01-24
Implement `--indent` and `--tab`


chutz:
`jq` has command line option to allow one to control how the output is indented, it would be useful if `jaq` supported similar options.

PR #153, opened 2024-01-24
Add `--indent` and `--tab` command line options

chutz:
Attempt to implement `--indent` and `--tab` command line options as suggested in https://github.com/01mf02/jaq/issues/152

Let me know if there is a better way to do this, I am a beginner at rustlang.


01mf02:
This is nice. Thanks a lot!

PR #158, opened 2024-02-04
Implement indices, index and rindex.

### LINKED ISSUE #154: Support JQ's index/rindex function
Jq has the [`index`](https://jqlang.github.io/jq/manual/#index-rindex) and `rindex` functions which don't seem to be supported in jaq.  Is there a specific reason for this?  Thank you for your time.

kklingenberg:
This PR implements `indices` and copies jq's derivative implementations of `index` and `rindex` according to the spec [here](https://jqlang.github.io/jq/manual/#indices). Note that jq has TODO's related to performance on said implementations, so perhaps you'd want to close this PR and wait for a performant version of these filters.

~Also, the behaviour of jq's `indices` varies slightly when dealing with strings. I reported/asked about it here: <https://github.com/jqlang/jq/issues/3030>~ (my mistake; jq solved the issue in 1.7)

Closes #154

### DIFF
--- a/jaq-std/src/std.jq
+++ b/jaq-std/src/std.jq
@@ -117,6 +117,28 @@ def any: any(.[]; .);
 def in(xs)    : . as $x | xs | has     ($x);
 def inside(xs): . as $x | xs | contains($x);
 
+# Indexing
+def indices($i):
+  def enumerate:
+    . as $thing |
+    if type == "string"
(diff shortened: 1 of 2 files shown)


01mf02:
Thanks a lot, this is some great work! Just for future reference, in tests, it is preferable to use `yields!` instead of `gives`, because it allows for much more compact tests with names.

PR #159, opened 2024-02-04
fix building jaq-interpret tests on 32-bit platforms

nekopsykose:
4999950000 overflows a usize (u32)


01mf02:
Thanks a lot for noticing this!

PR #171, opened 2024-03-26
Add Windows install (Scoop) instructions

IndyV:
Since I found this tool by looking for `jq` on [Scoop](https://scoop.sh/), I thought it could be nice to add install instructions on Windows.

Scoop is sort of like Homebrew for Windows. The manifest auto-updates to the latest release from this repository.
You can find the manifest for it [here](https://github.com/ScoopInstaller/Main/blob/master/bucket/jaq.json).


01mf02:
Thanks a lot!

PR #174, opened 2024-04-28
chore(core): export core functions based on enabled features

ssddOnTop:
The main idea behind this PR is that some applications are platform independent and they do not need std.

currently we have option for `jaq_core::core()` and `jaq_core::minimal()` but core needs all features to be enabled, there is no way that we can avoid excluding specific feature. 

This pr adds new function `feature_enabled` which chains with feature flags

### DIFF
--- a/jaq-core/src/lib.rs
+++ b/jaq-core/src/lib.rs
@@ -53,6 +53,27 @@ pub fn core() -> impl Iterator<Item = (String, usize, Native)> {
         .chain(run(TIME))
 }
 
+/// Feature enabled filters.
+pub fn feature_enabled() -> impl Iterator<Item = (String, usize, Native)> {
+    let mut feature = minimal();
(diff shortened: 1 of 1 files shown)


ssddOnTop:
I am not sure why we have core and minimal functions only but I have a 

proposal: make all feature specific implementations, `run` and `upd` functions public so we don't have to update several functions on addition of each feature

01mf02:
I know that you closed this issue, but it might still be worth to explain why I would not do it like this: In jaq, *feature flags should not change the output of any function of a library crate*. So enabling a feature flag should not make one function suddenly return more filters. However, it may *expose* new functions, which can be used in an application.

ssddOnTop:
I don't know if the feature flags are used anywhere else in the codebase, so I thought it would be good idea that people can only add feature they want and add functionality based on that

PR #175, opened 2024-04-28
chore(Val): impl (de)serializer

ssddOnTop:
This pr adds serializer and de-serializer for val

TODO: 
- [ ] add tests


ssddOnTop:
I know you are planning for ValT but I think it would be better implement serde for Val, it will eventually give more flexibly and maintainability to users

Issue #178, opened 2024-05-06
Build failure on 1.80 nightly and Windows


clin1234:
Running `cargo install  --all-features --locked --git https://github.com/01mf02/jaq` fails to compile:

```
error[E0635]: unknown feature `stdsimd`
  --> C:\Users\ךינשגכהד\.cargo\registry\src\index.crates.io-6f17d22bba15001f\ahash-0.8.6\src/lib.rs:99:42
   |
99 | #![cfg_attr(feature = "stdsimd", feature(stdsimd))]
   |                                          ^^^^^^^

For more information about this error, try `rustc --explain E0635`.
error: could not compile `ahash` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `ahash` (lib) due to 1 previous error
error: failed to compile `jaq v1.3.0 (https://github.com/01mf02/jaq#7baed83d)`, intermediate artifacts can be found at `C:\Users\0BAC~1\AppData\Local\Temp\cargo-install20BWIe`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

tkaitchuck/aHash#227 explains the issue.

01mf02:
This is a duplicate of #176.

01mf02:
Oops, I'm sorry, this is actually a different issue. Does [cd728e2](https://github.com/01mf02/jaq/commit/cd728e21761638a025078d183d5d89c37c7eb775) resolve it?

clin1234:
Compiles now.

Issue #176, opened 2024-04-30
Build failure for v1.3.0 on macos & linux, but building from latest main (1699a7d1fc7ede) works


derekfrye:
I just tried to compile the latest version with `cargo`, and maybe I'm doing something wrong but I can only get v1.2.0 to build. Here's what I tried. 

Note, if i either `cargo install jaq --version "1.2.0"` or I clone from git (`1699a7d1fc7ede` as of now) I can build either of those. So IDK..? I am pretty new to rust so maybe I'm missing a step?

```
$ uname -a
Darwin NUSLANYKC904450 23.4.0 Darwin Kernel Version 23.4.0: Fri Mar 15 00:19:22 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T8112 arm64
$ curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf | sh
$ rustc --version
rustc 1.77.2 (25ef9e3d8 2024-04-09)
$ cargo --version
cargo 1.77.2 (e52e36006 2024-03-26)
$ cargo install jaq
    Updating crates.io index
    Installing jaq v1.3.0
    Updating crates.io index
  Downloaded cc v1.0.96
  Downloaded 1 crate (76.6 KB) in 1.02s
   Compiling proc-macro2 v1.0.81
   ...
   Compiling jaq v1.3.0
error[E0599]: no variant or associated item named `Unset` found for enum `ariadne::Color` in the current scope
   --> /Users/a0761532/.cargo/registry/src/index.crates.io-6f17d22bba15001f/jaq-1.3.0/src/main.rs:476:17
    |
476 |         (Color::Unset, Color::Unset)
    |                 ^^^^^ variant or associated item not found in `Color`

error[E0599]: no variant or associated item named `Unset` found for enum `ariadne::Color` in the current scope
   --> /Users/a0761532/.cargo/registry/src/index.crates.io-6f17d22bba15001f/jaq-1.3.0/src/main.rs:476:31
    |
476 |         (Color::Unset, Color::Unset)
    |                               ^^^^^ variant or associated item not found in `Color`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `jaq` (bin "jaq") due to 2 previous errors
error: failed to compile `jaq v1.3.0`, intermediate artifacts can be found at `/var/folders/vf/9dvlj94d7d990822xvnpbqn40000gn/T/cargo-installa6qIVu`.
```
I get the same thing on my linux machine. 
[...]

PR #179, opened 2024-05-07
Update ahash 0.8.6 -> 0.8.11 for stdsimd feature and Color::Unset -> Color::Primary to reflect yansi v1.0.1

aretrosen:
Closes https://github.com/01mf02/jaq/issues/178 and https://github.com/01mf02/jaq/issues/176.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "ahash"
-version = "0.8.6"
+version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


01mf02:
Hi @aretrosen, thanks for your PR!
Unfortunately, ahash 0.8.8 increases the minimal Rust version to 1.72, whereas currently, jaq builds on at least Rust 1.64. According to https://github.com/tkaitchuck/aHash/issues/227, ahash 0.8.7 is sufficient to solve the issue and does not increase MSRV, so I've made the change myself in cd728e2.
The ariadne issue has to be corrected on the ariadne side. In order to prevent unwanted side effects, I'll stick with the current ariadne for now.

PR #181, opened 2024-05-12
Use `alloc::rc::Rc::try_unwrap` from stdlib over custom implementation

SamuelMarks:


01mf02:
I'm sorry, but this change would raise MSRV from 1.64 to 1.76, which is not warranted by any added functionality.

I will perform this change once there is a strong benefit gained from increasing MSRV to >= 1.76.

PR #193, opened 2024-07-07
chore(deps): bump mimalloc dependency

tranzystorekk:
The newest mimalloc version contains a [patch](https://github.com/purpleprotocol/mimalloc_rust/commit/446fc231151cfab485b0ac7d5178baf797799dda) for building on the armv6 target, present for example on Void Linux.


01mf02:
Thanks!

PR #199, opened 2024-07-30
Use `chrono` instead of `time`.

01mf02:
The crate `time` does not build on Rust 1.80 for versions smaller than 0.3.36: https://github.com/time-rs/time/issues/681
Apart from the fact that I do not like packages stopping to build when *updating* the compiler, updating time to 0.3.36 would require raising MSRV to 1.67, which I do not like.

I found that the `chrono` crate has a lower MSRV, it works fine on all compiler versions I tried, and using it is considerably simpler for my use case, so this PR replaces `time` with `chrono`. 
A small side effect of this is that jaq outputs fractional seconds with 6 digits instead of 9 before.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,6 +36,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "autocfg"
+version = "1.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0c4b4d0bd25bd0b74681c0ad21497610ce1b7c91b1022cd21c80c6fbdd9476b0"
+
+[[package]]
 name = "base64"
 version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -64,6 +70,15 @@ name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
+
+[[package]]
+name = "chrono"
+version = "0.4.38"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a21f936df1771bf62b77f047b726c4625ff2e8aa607c01ec06e5a05bd8463401"
+dependencies = [
+ "num-traits",
+]
 
 [[package]]
 name = "clap"
@@ -240,6 +255,7 @@ version = "1.5.0"
 dependencies = [
  "aho-corasick",
  "base64",
+ "chrono",
  "hifijson",
  "jaq-interpret",
  "jaq-syn",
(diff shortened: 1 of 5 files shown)


jhpratt:
> Apart from the fact that I do not like packages stopping to build when updating the compiler

This is not the fault of `time`.

> updating time to 0.3.36 would require raising MSRV to 1.67, which I do not like.

Why do you need an MSRV that is more than 18 months old? Very few crates have policies that extend that far back (with `chrono` not having a policy at all, mind you).

01mf02:
> > Apart from the fact that I do not like packages stopping to build when updating the compiler
> 
> This is not the fault of `time`.

I'm sorry, I was indeed under the impression that this was the fault of `time`. Rereading your post (https://github.com/time-rs/time/issues/681#issuecomment-2143130755), I see that this does not seem to be the case after all. I hope that you accept my apology.
Still, I do not understand how adding a new trait implementation to the standard library could break your code. I thought that Rust's coherence rules prevent such breakage? If you can enlighten me on this point, I would be very interested.

> > updating time to 0.3.36 would require raising MSRV to 1.67, which I do not like.
> 
> Why do you need an MSRV that is more than 18 months old? Very few crates have policies that extend that far back (with `chrono` not having a policy at all, mind you).

A colleague of mine who was not into Rust at that time once tried to install one of my Rust programs on Debian and failed because his system-installed compiler was too old. I just checked now, and the [rustc shipped with stable Debian is 1.63](https://packages.debian.org/bookworm/rust/rustc). So I'm trying to preserve MSRV as long as it is not too much effort.

jhpratt:
> Still, I do not understand how adding a new trait implementation to the standard library could break your code. I thought that Rust's coherence rules prevent such breakage? If you can enlighten me on this point, I would be very interested.

Essentially, the code I had written relied on the fact that the return type of `.into()` could be inferred. Type inference is something explicitly permitted as breakage. With the addition of `IntoIterator for Box<[T]>`, this was no longer the case. This specific instance of breakage was caught months ago, which is why a patch was merged and released in April.

> A colleague of mine who was not into Rust at that time once tried to install one of my Rust programs on Debian and failed because his system-installed compiler was too old.

I honestly don't understand why they even bother shipping a version so old (it was released two full years ago). `rustup` is _always_ the recommended way to do anything related to Rust, in my opinion.

01mf02:
> I honestly don't understand why they even bother shipping a version so old (it was released two full years ago). `rustup` is _always_ the recommended way to do anything related to Rust, in my opinion.

I also use `rustup`, but from my own experience with people who are just trying Rust, the majority first tries to install a Rust toolchain using their package manager. And that's only one case of people that I know about who might be stuck with an older Rust toolchain. There might be others. Keeping MSRV low means to cater to the needs of these.

01mf02:
> Essentially, the code I had written relied on the fact that the return type of `.into()` could be inferred. Type inference is something explicitly permitted as breakage. With the addition of `IntoIterator for Box<[T]>`, this was no longer the case. This specific instance of breakage was caught months ago, which is why a patch was merged and released in April.

Thanks for your explanation, @jhpratt!

Issue #200, opened 2024-07-31
Why `ParseCtx.compile` doesn't return a Result?


lquerel:
The implementation of `ParseCtx.compile(jaq_syn::Main)` is currently error-prone because, if internally the compilation fails, the `compile` method returns a default Filter (i.e., `.`) and the public field `ParseCtx.errs` is populated with the errors. This design makes the usage of `ParseCtx.compile()` a bit error-prone as it's not obvious that the field `errs` must be checked. So, my question is, why not return a `Result<Filter, Vec<Error>>` instead of `Filter` and get rid of the field `errs` in `ParseCtx`?

01mf02:
I agree with you totally. I'm currently working on jaq 2.0, and I plan to return `Result`s for functions like the one you mentioned. This will also cut down on other boilerplate as @Andrew15-5 suggested in #188.

PR #202, opened 2024-08-14
Rewrite compiler

### LINKED ISSUE #157: Discrepancy with jq and gojq: subfunction after variable declaration
Using jaq.jq below:
```
$ jq -n -f jaq.jq
1
2

$ goq -n -f jaq.jq
1
2

$ jaq -n -f jaq.jq
Error: Unexpected token while parsing definition, expected [, if, try, {, .., -, ., (, "
   ╭─[<unknown>:5:5]
   │
 5 │   | def a: 2 * $s;
   │     ──┬─  
   │       ╰─── Unexpected token def
───╯
```

---
jaq.jq
```
def example:
  . as $s
  | def a: 2 * $s;
  ($s, a);

1 | example
```

### LINKED ISSUE #188: Consider creating easy-to-use library version/crate
Hi. When I was searching for the Rust-native `jq` library (several months ago), I was surprised that I only found CLIs. But in a desperate attempt, I tried to "reverse-engineer" the CLI. I don't remember if jaq was the only viable option or not, but I chose it. It was extremely hard as there was almost no documentation for my case, so I started bisecting the CLI crate. In the end, the "boilerplate" became pretty huge. It definitely can be reduced if a dedicated library crate is used, but there is none.

<details><summary>the boilerplate</summary>

```rs
fn apply_json_filter<T>(json: serde_json::Value, filter: &str) -> Result<T>
where
    T: DeserializeOwned,
{
    let mut defs = ParseCtx::new(vec![]);
    defs.insert_natives(jaq_core::core());
    defs.insert_defs(jaq_std::std());

    let (main_filter, errs) = jaq_parse::parse(filter, jaq_parse::main());
    if !errs.is_empty() {
        return Err(make_error_message_from_vec(
            "There were some errors parsing the filter command:\n",
            errs,
            |e| format!("{}\n", e),
        ));
    }
    if main_filter.is_none() {
        return Err(Error::msg(
            "Main filter is `None` For some reason, even though there were no errors.",
        ));
    }

    let finished_filter = defs.compile(main_filter.unwrap());
    if !defs.errs.is_empty() {
        return Err(make_error_message_from_vec(
            "There were some errors compiling the main filter:\n",
            defs.errs,
            |e| format!("Error parsing characters in range: {:?}\n", e.1),
        ));
    }

    let inputs = RcIter::new(core::iter::empty());

    let out: Vec<_> = finished_filter
        .run((Ctx::new([], &inputs), Val::from(json)))
        .collect();

    if out.iter().any(|x| x.is_err()) {
        return Err(make_error_message_from_vec(
            "Failed to filter JSON object(s):\n",
            out.iter().filter(|x| x.is_err()).collect(),
            |e| format!("{:?}\n", e),
        ));
    }
    let objects: Vec<_> = out
        .into_iter()
        .map(|x| serde_json::Value::from(x.unwrap()))
        .collect();

    let filtered_json = if objects.len() == 1 {
        objects.into_iter().next().unwrap()
    } else {
        serde_json::Value::Array(objects)
    };

    serde_json::from_value(filtered_json).map_err(Error::from)
}
```

</details>

My implementation required at least 4 separate crates: `jaq-interpret`, `jaq-parse`, `jaq-core`, `jaq-std` (in addition to the `serde` and `serde_json`). I really hope that a library crate can be created that can provide good defaults in addition to some configuration (for optimizing performance and dependency tree). I also noticed that I'm not the first one who faced this problem and created an issue. Perhaps my function can be a starting point.

01mf02:
This PR adds support for many fundamental new features, such as:

* modules (`include "mod"; import "mod" as mod;`)
* imported data files (`import "data" as $data;`)
* label...break expressions (`label $x | break $x`)
* generalised nested definitions (`1 + def a: 2; a`; closes #157).
* `//=` operator (inspired by #142, thanks to @bb010g)

Where the old compiler performed three different passes for compilation (HIR -> LIR -> MIR), the new compiler only takes a single pass. This should benefit performance.
It also comes with a completely rewritten API to compilation, hopefully simplifying its usage (closes #188, #200).

Compared to jq and gojq, circular dependencies between modules are detected and warned about.


pkoppstein:
Sorry if I'm jumping the gun, but using the latest (Aug 14) version of jaq:
```
$ git pull
Already up to date.

$ git rev-parse --short HEAD
202736c

$ ls -l target/release/jaq
-rwxr-xr-x  1 peter  staff  3482416 Aug 15 00:08 target/release/jaq

$ target/release/jaq -n 'include "bitwise" {"search": "."};'
Error: expected ;
  ╭─[<inline>]
  │
1 │ include "bitwise" {"search": "."};
  ┆                   ───────┬─────── 
  ┆                          │        
  ┆                          ╰───────── unexpected token
──╯
```
Incidentally, it would be very helpful if `jaq --version' gave
the (short) git commit id when jaq is compiled locally.

Thank you!

01mf02:
Hello Peter, I added support for parsing metadata at the end of include/import directives in ee4fa58e42d0e3b6015217bcd29cf7fb0c320ac8. However, the documentation of what should happen with that module metadata is so vague that I do not feel able to implement this stuff in more detail. Therefore, jaq currently just ignores the metadata. If you can explain to me in more detail how jq uses this metadata (in particular how `search` works), then I might be able to support that use case. Going as far as supporting `modulemeta` is probably out of reach for jaq 2.0.

And while I agree with you on the `jaq --version` issue, we've been through this in #40, and it's unfortunately quite hard to do this.

pkoppstein:
@01mf02 wrote:
>  If you can explain to me in more detail how jq uses this metadata (in particular how search works)…

Indeed, the “search” key is the one that matters the most, by far.  I think it’s conceptually quite straightforward - it just overrides everything else, which makes sense as it’s the most specific possible indicator. There’s also nothing complicated about the logic, so I would expect (and certainly hope) it would be straightforward to implement. Please let me know if there’s anything else I might be able to help clarify. Meanwhile, thank you!

01mf02:
So let's suppose that we have a module "x/toto.jq" that contains the statement `include "foo" {search: "bar"};`. Does that now mean that the file "foo.jq" is searched in the path "./bar", or in "x/bar"? Are files that are included from "foo.jq" also searched in "./bar" (or "x/bar")?

pkoppstein:
My apologies. I was writing of the most important case (at least in my view) and have now clarified the text to emphasize what the manual says:  

> The search key in the metadata, if present, should have a string or array value (array of strings); this is the search path to be prefixed to the top-level search path.

This means that if the “search key” is a string, $s (such as “.”) and the module being imported is named "foo", and if the file "foo.jq" exists in the directory specified by $s (relative to the pwd), then it will be used.

01mf02:
Thanks for the clarification, @pkoppstein. It leads me to another question: Is `include "<foo>/bar";` always equivalent to `include "bar" {search: <foo>};`? If not, can you give me an example where the two differ?
Also, does `{search: "."}` actually have any effect?

pkoppstein:
@01mf02 asked:

> Is include "<foo>/bar"; always equivalent to include "bar" {search: <foo>}; (for any string <bar>)? 

> Also, does {search: "."} actually have any effect?

In summary, the answers are: NO!!! and YES!!!

Let's go back to basics.  Using the terminology of the jq manual, the syntax for `include` is:
```
 include RelativePathString [<metadata>];   
```

Here, `RelativePathString` must be a *relative* pathname.  (Whether using the C or Go implementations, an error will be raised is an absolute pathname is specified.). This relative pathname is resolved by searching (left-to-right) through the current "list of jq library paths", which in turn is determined by (a) the "search" key if specified; (b) the paths specified by the -L command-line option (if specified); and (c) various defaults.

So, for example, specifying `{search: "."}` places the pwd at the beginning of the "jq library paths" list.

(In the C implementation, you can grep for JQ_LIBRARY_PATH in the src/ files.)

pkoppstein:
@01mf02 wrote:
> ... jq -n 'include "a/b"; 0' ...

Using your setup, gojq works as expected.  So it looks like the C implementation has a bug; alas, that wouldn't surprise me at all. Fortunately, since we have the C and Go implementations as well as the documentation, it's usually possible to separate the wheat from the chaff.

pkoppstein:
> If the "search" key is specified, should the paths specified by the -L command-line option be considered? 

Yes.  The documentation says:

> The search key in the metadata, if present, should have a string or array value (array of strings); this is the search path to be prefixed to the top-level search path.

Here "prefixed" effectively means "prepended to the array".   See my remarks about "list of jq library paths" above.
I think it's safe to say that JQ_LIBRARY_PATH is modeled after LD_LIBRARY_PATH and similar *_LIBRARY_PATH offerings.

01mf02:
Thanks for the explanations, @pkoppstein. I have implemented the rules in #207. If you want to try it, your feedback would be most welcome. I just diverge from jq's rules by not searching `["~/.jq", "$ORIGIN/../lib/jq", "$ORIGIN/../lib"]` by default, as this part would add quite a number of additional dependencies to jaq.

pkoppstein:
Using commit 141202c, I've verified that some important cases are handled properly, so first let me thank you for this accomplishment.  Two quick observations:

1) Did you overlook `{search: "~/"}`?

```
$ cat ~/a.jq
def a: "Here";

$ (cd $(dirname $JAQ); git log --oneline -1 )
141202c (HEAD -> load-meta, origin/load-meta) Make read function a member of Import.

$ cd ~/jq/modules
$ jq -n -L. 'include "a" {search: "~/"}; a'
"Here"

$ $JAQ -n -L. 'include "a" {search: "~/"}; a'
Error: expected nothing
  ╭─[./a.jq]
  │
5 │ a
  ┆ ┬
  ┆ │
  ┆ ╰─ unexpected token
──╯
$
```
(2) You mentioned deliberately not supporting the default library path `["~/.jq", "$ORIGIN/../lib/jq", "$ORIGIN/../lib"]`

Quite a lot of thought went into this default, which is not to say the situation is ideal, but having a default is handy, and having ~/.jq as the first default makes sense as modules are typically placed therein.  Also, supporting ~/.jq now would by no means preclude adding the $ORIGIN-based directories at some later time.

pkoppstein:
@01mf02 - I'm no rustacean but I believe the following would provide sufficient portability:
```
use std::env;
use std::path::{PathBuf};

fn home_dir() -> Option<PathBuf> {
    if cfg!(target_os = "windows") {
        env::var_os("USERPROFILE").map(PathBuf::from)
    } else {
        env::var_os("HOME").map(PathBuf::from)
    }
}
```

jaq already uses 
```
std::{env, io, process, str};
std::path::{Path, PathBuf};
```
so I don't think there's an issue there. 

Regarding your point about `{search: "~/jq"}` and the like, please note that that usage would typically be private, but even if it were shared, the consequences would be minimal as the "search" path is just prepended to the usual path.

pkoppstein:
To capture the $ORIGIN I believe the following illustrates a sufficiently portable and minimally demanding approach:
```
use std::env;

fn main() {
    // Get the full path of the current executable
    let current_exe_path = env::current_exe()
        .expect("Failed to get the path of the current executable");

    // Get the directory in which the executable resides
    let current_dir = current_exe_path.parent()
        .expect("Failed to get the directory containing the executable");

    println!("Executable is located in: {}", current_dir.display());
}
```

01mf02:
Concerning `~`: While the approach you highlighted seems reasonable on the first look, I wanted to double-check and looked at the `dirs` crate. [Its home directory logic](https://docs.rs/crate/dirs-sys/0.4.1/source/src/lib.rs) is significantly more complex, because it also handles more platforms (Redox OS, macOS) --- all platforms that I do not use myself, but that I would like to support. When I try to add `dirs` to my dependencies, the result looks like this:

~~~
$ cargo add dirs
    Updating crates.io index
      Adding dirs v5.0.1 to dependencies
    Updating crates.io index
     Locking 17 packages to latest compatible versions
      Adding bitflags v2.6.0
      Adding dirs v5.0.1
      Adding dirs-sys v0.4.1
      Adding libredox v0.1.3
      Adding option-ext v0.2.0
      Adding redox_users v0.4.6
      Adding thiserror v1.0.55 (latest: v1.0.63)
      Adding thiserror-impl v1.0.55 (latest: v1.0.63)
      Adding windows-sys v0.48.0 (latest: v0.59.0)
      Adding windows-targets v0.48.5 (latest: v0.52.6)
      Adding windows_aarch64_gnullvm v0.48.5 (latest: v0.52.6)
      Adding windows_aarch64_msvc v0.48.5 (latest: v0.52.6)
      Adding windows_i686_gnu v0.48.5 (latest: v0.52.6)
      Adding windows_i686_msvc v0.48.5 (latest: v0.52.6)
      Adding windows_x86_64_gnu v0.48.5 (latest: v0.52.6)
      Adding windows_x86_64_gnullvm v0.48.5 (latest: v0.52.6)
      Adding windows_x86_64_msvc v0.48.5 (latest: v0.52.6)
~~~

This is the list of dependencies that are added. And because `~` expansion happens during compilation of jq modules, this adds 17 dependencies to the central crate of jaq, `jaq-core`, which currently only has three dependencies (all of which are platform-agnostic). This is one main reason why I'm reluctant to perform this change.

01mf02:
Your $ORIGIN code looks fine, on the other hand. However, I'd like to support either both `~` and `ORIGIN`, or none of both.

01mf02:
> Regarding your point about {search: "~/jq"} and the like, please note that that usage would typically be private, but even if it were shared, the consequences would be minimal as the "search" path is just prepended to the usual path.

If the usage would typically be private, couldn't users who put `{search: "~/jq"}` into their jq files simply use `jaq -L ~/jq` to satisfy their needs?

pkoppstein:
@01mf02 wrote:
> couldn't users who put {search: "~/jq"} into their jq files simply use jaq -L ~/jq

Sure, but there's a delicate dance going on here.  For example, forcing users to use -L makes things difficult if they also want to take advantage of the $ORIGIN conventions.

> cargo add dirs

I'm not sure what the advantages of adding that crate would be, but it does seem to me that it would be
acceptable, at least for now, to rely on std::env and std::path, with the explanation that `~` expansion is only supported on Windows and wherever "HOME" is set and available via env_var_os.  And vastly preferable to not providing a default at all, which I understand to be what you have in mind.

wader:
Could it make sense to make it possible to optionally dependency inject a "get home dir"-function somehow? then it would be up to a jaq crate user to do whatever they want like depend on the dirs crate?

01mf02:
@pkoppstein, @wader, given your interest in this functionality, I decided to implement it after all in c3f9259e4c5d5921a14cbc2143349119a761bcb5.

pkoppstein:
@01mf02 - Greatly appreciated!  Would I be correct in thinking that you opted for the "simplistic" approach to avoid additional crates?  Anyway, thank you.

01mf02:
@pkoppstein: Yes, I opted for the "simplistic" approach that you suggested.

PR #205, opened 2024-08-23
docs: Add IPA pronunciation

thecaralice:
IPA pronunciation is more straightforward than "like <word>", especially for non-native English speakers who do not know how to pronounce every French word in English.


01mf02:
Thank you!

PR #210, opened 2024-09-10
Fix escaping of unicode characters in strings

### LINKED ISSUE #209: jaq outputs invalid utf8 escaped json
I'm not sure if jaq is outputting invalid utf8, or if jq is too liberal in what it accepts. In any case this is easy to reproduce:

Steps:
1. Have a json string with unicode codepoints
2. Pass it through jaq

Expected:
1. jaq outputs the unicode characters in json

Actual:
1. jaq tries to escape the unicode characters

```javsacript
// foo.js
a = "\u{200b}banana \u{200e}man"
console.log(JSON.stringify({a}))
```

```bash
# this outputs the string with the unicode characters (but they're not visible)
❯ node foo.js 
{"a":"​banana ‎man"}

# jaq tries to escape the unicode characters, but this is invalid json
❯ node foo.js | jaq
{
  "a": "\u{200b}banana \u{200e}man"
}

# jq outputs the unicode characters (but they're not visible)
❯ node foo.js | jq
{
  "a": "​banana ‎man"
}

# we can see jaq is producing invalid json
❯ node foo.js | jaq | jaq
Error: failed to parse: invalid hexadecimal sequence

# but jq produces valid json
❯ node foo.js | jq | jaq 
{
  "a": "\u{200b}banana \u{200e}man"
}

❯ node foo.js | jaq | jq 
jq: parse error: Invalid characters in \uXXXX escape at line 2, column 35

```

ibash:
Attempting to fix #209

Previously jaq would output strings with unicode escaped with rust debug format like:

```
{
  "a": "\u{200b}foo \u{200e}bar"
}
```

This is invalid json. Instead those unicode characters should be output directly.

This commit uses serde_json to serialize strings instead which escapes them correctly for json.

Unfortunately this change makes the serde_json feature required, which might not be desired for this project.

### DIFF
--- a/jaq-json/src/lib.rs
+++ b/jaq-json/src/lib.rs
@@ -9,6 +9,9 @@ use core::fmt::{self, Debug};
 use jaq_core::{load, ops, path, Exn, Native, RunPtr};
 use jaq_std::{once_with, ow, run, unary, v, Filter};
 
+#[cfg(feature = "serde_json")]
+use serde_json;
+
 #[cfg(feature = "hifijson")]
 use hifijson::{LexAlloc, Token};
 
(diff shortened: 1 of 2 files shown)


01mf02:
Closing as explained in #209.

PR #212, opened 2024-09-23
Add `trim/0`, `ltrim/0` and `rtrim/0`

wader:
Trims leading and trailing whitespace.

Was added to jq in https://github.com/jqlang/jq/pull/3056

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,7 @@ Here is an overview that summarises:
 - [x] String <-> integers (`explode`, `implode`)
 - [x] String normalisation (`ascii_downcase`, `ascii_upcase`)
 - [x] String prefix/postfix (`startswith`, `endswith`, `ltrimstr`, `rtrimstr`)
+- [x] String whitespace trimming (`trim`, `ltrim`, `rtrim`)
 - [x] String splitting (`split("foo")`)
 - [x] Array filters (`reverse`, `sort`, `sort_by(-.)`, `group_by`, `min_by`, `max_by`)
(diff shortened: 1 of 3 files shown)


wader:
In the jq version of this there is an optimization to not create a new string if there is nothing to trim https://github.com/jqlang/jq/pull/3056/files#diff-bd792def7de6d4b8d70ddc844eabe52f73c264904e263580215219b669a2ca9cR1239-R1242 i tried to follow the rust code but not sure i understood fully how it would behave

wader:
Thanks 🙏  and nice you added the optimisation, i haven't researched it but i suspect that most calls to trim will not trim anything. Maybe there are more string functions that could benefit of the same optimisation?

01mf02:
I just checked that, and `ltrimstr` as well as `rtrimstr` already perform this optimisation.
`escape_csv` and `escape_sh` are unfortunately no candidates for this optimisation due to the way that [`str::replace`](https://doc.rust-lang.org/std/primitive.str.html#method.replace) works. The same goes for `ascii_downcase` and `ascii_upcase`.

wader:
👍 now also remembered that i forgot to add tests for some different unicode whitespace codepoints, there is a bunch of different ones, ex a test like this https://github.com/jqlang/jq/blob/master/tests/jq.test#L1363-L1367. I did a quick check and it seems like jq, gojq and jaq all agree on what is whitespace now, ex javascript does not see `U+0085 NEXT LINE` as whitespace.

PR #215, opened 2024-09-24
Update on surprising jq behaviours

wader:
- `[[]] | implode` crash was fixed in 1.7 by https://github.com/jqlang/jq/pull/2646
- `[limit(0; 1, 2, 3)]` yielding 1,2 instead of nothing fixed in 1.7 by https://github.com/jqlang/jq/issues/2316
- `limit/2` will throw error on negative limit once https://github.com/jqlang/jq/pull/3181 is merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,13 +19,6 @@ jaq focuses on three goals:
   <details><summary>Examples of surprising jq behaviour</summary>
 
   * `nan > nan` is false, while `nan < nan` is true.
-  * `[[]] | implode` crashes jq, and this was not fixed at the time of writing despite
-    [being known since five years](https://github.com/jqlang/jq/issues/1160).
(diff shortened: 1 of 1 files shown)


wader:
The NaN compare stuff is a bit weird, maybe would be ok to do a breaking change for that? or could there be some indirect use of NaN compare that would break?

About the performance issues: i think most of them got resolved by https://github.com/jqlang/jq/pull/1834

01mf02:
Hi @wader, thanks for your PR! Your push motivated me to also remove a few other claims from the README that referred to jq 1.6 and which were not longer true in jq 1.7. So thanks for that. :)

wader:
@01mf02 👍 btw i'm slowly reading your "A formal specification of the jq language" paper and hope to get to review your jq docs improvements, been a bit busy with non-jq stuff

PR #218, opened 2024-10-15
Patterns in variable bindings.

01mf02:
This PR adds support for patterns in variable bindings, e.g. `. as {a: [$x, {("b", "c"): $y, $z}]} | $x, $y, $z`.
This also works in `foreach` and `reduce`.

Compared to jq, this also allows indexing with integers, e.g. `{(0): $x, (1): $y}`, which is equivalent to `[$x, $y]`. This can be useful if you wish to destructure arrays, but not necessarily the starting items.

However, jaq does not support the (undocumented) syntax `{$x: ...}`; for example, the filter `{x: [1]} as {$x: [$y]} | $x, $y` this yields `[1]` and `1` in jq.

### DIFF
--- a/jaq-core/src/box_iter.rs
+++ b/jaq-core/src/box_iter.rs
@@ -1,10 +1,24 @@
+//! Boxed iterators.
+
 use alloc::boxed::Box;
 
+/// A boxed iterator.
 pub type BoxIter<'a, T> = Box<dyn Iterator<Item = T> + 'a>;
+
+/// A boxed iterator over `Result`s.
+pub type Results<'a, T, E> = BoxIter<'a, Result<T, E>>;
 
 /// Return a boxed iterator that yields a single element.
 pub fn box_once<'a, T: 'a>(x: T) -> BoxIter<'a, T> {
     Box::new(core::iter::once(x))
+}
+
+/// If `x` is an `Err`, return it as iterator, else apply `f` to `x` and return its output.
+pub fn then<'a, T, U: 'a, E: 'a>(
+    x: Result<T, E>,
+    f: impl FnOnce(T) -> Results<'a, U, E>,
+) -> Results<'a, U, E> {
+    x.map_or_else(|e| box_once(Err(e)), f)
 }
 
 /// For every element `y` returned by `l`, return the output of `r(y, x)`.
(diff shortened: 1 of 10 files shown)


wader:
Impressive work! and i had no idea jq supported combining key/value pattern like`{$x: [$y]}`. Index based patter for arrays seems nice addition, wonder how much work it would be to get into jq. I've also wished sometimes for some kind of rest-syntax like `[$x, ...$xs]` and `{$k, ...$obj_with_all_other_keys}`

01mf02:
Regarding `{$x: pattern}`, I have not implemented this because of a few reasons:

* `{$x: pattern}` is _not_ the same as `{($x): pattern}`. I can't think of any other place in jq where `($x)` cannot be substituted by `$x`, so this is some really surprising behaviour. If you only know the `{(f): pattern}` syntax and are confronted with `{$x: pattern}`, you may be up for a pretty big surprise.
* `{$x: pattern}` discriminates keys that are not "identifier-like"; that is, this only works for fields like `"x"`, but not for `"here-be-dragons"`. (The same could be said of course about `.x` and `{$x}`, but these can be transformed quite simply into equivalent constructs that work with non-identifier-like keys, such as `.["x"]` and `{"x": $x}`. There is no easy transformation for `{$x: pattern}`, however. This can be really cumbersome if you have used `{$x: pattern}`, perhaps somewhere inside another pattern, then your input format changes from `{"x": ...}` to `{"here-be-dragons: ...}`. Have fun rewriting your pattern, especially if you use it inside of `reduce` or `foreach`.)

01mf02:
There is also yet another reason why I believe that `{$x: pattern}` is not a good idea: It provides useful functionality for objects, but there is no equivalent functionality for arrays, which would be equally useful. In particular, we might want to destructure `[{a: $x}]` *and* also bind a variable to the first element of the input array, namely to the object that `{a: $x}` is supposed to match. In jq's destructuring today, this is not possible.

We could come up with some syntax such as

    [$o as {a: $x}]

to solve this problem. In particular, wherever we can write a pattern, we should be able to write `$x as PATTERN`, in order to say that `$x` should bind to the whole value that is matched by `PATTERN`.

That would subsume the `{$x: PATTERN}` syntax, namely by writing:

    {x: $x as PATTERN}

Speaking as somebody who just implemented destructuring in jaq a few days ago, I think that such a syntax should be quite feasible to implement. And speaking as somebody who is trying to rewrite the jq manual, I think that such a syntax should be also relatively easy to describe, teach, and understand.

01mf02:
> I've also wished sometimes for some kind of rest-syntax like `[$x, ...$xs]` and `{$k, ...$obj_with_all_other_keys}`

While I also think that such syntax could be nice, I think that especially the second idea, namely the `$obj_with_all_other_keys`, could be a real pain to implement. The problem is that there is no elementary operation in jq that subtracts keys from objects, which would be useful in this case.
The first idea would be mostly subsumed by `{(0): $x, (-1): $xs}` that I implemented in jaq.

PR #219, opened 2024-10-21
Fix floats not getting converted to ints after rounding + more strict equality in tests

null-dev:
Currently `jaq`'s handling of floats differs from the [behavior specified in the docs](https://github.com/01mf02/jaq?tab=readme-ov-file#numbers):
```
❯ jaq --version; jaq -n '1.2 | [floor, round, ceil]'
jaq 1.6.0
[
  1.0,
  1.0,
  2.0
]
```

`git bisect` shows that this started happening after this commit: https://github.com/01mf02/jaq/commit/dfc6fc0f352bd704f1e26a6f65d168cdc42de89d

This PR fixes the behavior so floats are converted to ints again after rounding:
```
❯ cargo run -- -n '1.2 | [floor, round, ceil]'
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/jaq -n '1.2 | [floor, round, ceil]'`
[
  1,
  1,
  2
]
```

---

I've also tried to tighten up the equality check used during tests so it fails:
- If both values are of different types
- Or if the fields in both values are in a different order

This was a little complicated, and I'm not sure if this is even something you want, if it's worth the extra complexity or if the implementation is to your liking. Feedback is welcome, or you can also just close this PR and cherry-pick the float fix yourself if you don't want the test equality changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -511,6 +511,7 @@ version = "1.0.108"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3d1c7e3eac408d115102c4c24ad393e0821bb3a5df4d506a80f85f7a742a526b"
 dependencies = [
+ "indexmap",
  "itoa",
  "ryu",
  "serde",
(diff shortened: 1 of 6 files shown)


01mf02:
Hi @null-dev, thanks a lot for your PR and for spotting this!

Indeed, I think that while your `StrictEq` approach has merits in tests, it does not correspond to an existing notion of equality in jq. Therefore, I have created my own commit that cherry-picks your rounding commit only. However, I'm now testing whether the output of `round` & friends is integer, by checking that `[0, 1][1 | round]` returns 1 (instead of an error, as it would have before your fix).

Thanks again!

PR #221, opened 2024-10-30
Allow empty regex matches unless using ignore empty flag

wader:
Fixes `"" | test("")` being false.


01mf02:
Hi @wader, thanks for your PR. While reading through my old code, I found it quite difficult to understand, so I just rewrote it with iterators. Your tests, which I included in a5f868091f2105828be617fde0e83ae88cda5ed3, have been very useful in testing the implementation.

wader:
👍 btw the regex crate does not have a codepoint mode? always work with byte offsets?

01mf02:
> 👍 btw the regex crate does not have a codepoint mode? always work with byte offsets?

It seems so, yes. In Rust, nearly every string/character operation uses byte indices.

PR #229, opened 2024-11-11
Make reminder be integer reminder

wader:
This is same as jq modulo (hoh) error message:
```
$ jq -n 'def f: -2, -1, 0, 2.1, 3, 4000000001; [f as $a | f as $b | try ($a % $b) catch .]'
[
  0,
  0,
  "number (-2) and number (0) cannot be divided (remainder) because the divisor is zero",
  0,
  -2,
  -2,
  -1,
  0,
  "number (-1) and number (0) cannot be divided (remainder) because the divisor is zero",
  -1,
  -1,
  -1,
  0,
  0,
  "number (0) and number (0) cannot be divided (remainder) because the divisor is zero",
  0,
  0,
  0,
  0,
  0,
  "number (2.1) and number (0) cannot be divided (remainder) because the divisor is zero",
  0,
  2,
  2,
  1,
  0,
  "number (3) and number (0) cannot be divided (remainder) because the divisor is zero",
  1,
  0,
  3,
  1,
  0,
  "number (4000000001) and number (0) cannot be divided (remainder) because the divisor is zero",
  1,
  2,
  0
]
```


jaq-json/src/lib.rs
```diff
@@ -765,9 +765,14 @@ impl core::ops::Div for Val {
...
     fn rem(self, rhs: Self) -> Self::Output {
-        use Val::Int;
+        use Val::{Float, Int, Num};
         match (self, rhs) {
             (Int(x), Int(y)) if y != 0 => Ok(Int(x % y)),
+            (Float(f), Int(i)) if i != 0 => Ok(Int(f as isize % i)),
+            (Int(i), Float(f)) if f != 0.0 => Ok(Int(i % f as isize)),
+            (Float(x), Float(y)) if y != 0.0 => Ok(Int(x as isize % y as isize)),
+            (Num(n), r) => Self::from_dec_str(&n) % r,
+            (l, Num(n)) => l % Self::from_dec_str(&n),
```
wader, line 775: As i understand these will end up in the operator overload `rem` again?

wader:
Seems CI failure is same as what main branch is failing for also

01mf02:
In accordance with what I wrote in https://github.com/01mf02/jaq/issues/232#issuecomment-2469869339, I think that the  behaviour in this PR is not desirable, because it silently loses information about the decimal points. I'm sorry. I hope you understand.

wader:
No worries :) and the reason why jaq does not do non-integer reminder is because jq lacks it? i wonder if jaq's README or documentation could benefit from some notes and gotchas about how to do integer reminder etc, ex using ceil etc?

01mf02:
Hmmm ... we could actually implement non-integer remainders quite easily, because we can already do `l % r` in Rust for floating-point numbers `l` and `r`, see: https://doc.rust-lang.org/std/primitive.f64.html#impl-Rem-for-f64
That means that when we calculate `l % r` and either `l` or `r` are float, then the other argument should be converted to a float as well and floating-point remainder should be calculated.
I think that if you would adapt your PR that way, I would accept it.

> i wonder if jaq's README or documentation could benefit from some notes and gotchas about how to do integer reminder etc, ex using floor etc?

I thought that the "Numbers" section of the README did already contain this information. If you have a specific text in mind, I'm all for a little PR. :)

wader:
Switch float, but i'm not sure i really follow how reminder and floats work :) have to read more about it
```
$ cargo run -- -n -- '-2 / 2.1, -2 % 2.1'
-0.9523809523809523
-2.0
```

jaq-json/tests/funs.rs
```diff
@@ -77,3 +77,46 @@ fn tojson() {
...
+        0.10000000000000009,
+        0.10000000000000009,
+        "cannot calculate 2.1 % 0",
+        0.0,
+        2.1,
+        2.1,
+        1,
+        0,
+        "cannot calculate 3 % 0",
+        0.8999999999999999,
```
wader, line 112: Bad idea portabilitywise to end up with these kinds of values in tests?

01mf02: Yes. One trick to make this portable is to use something like `. * 1000 | round`, potentially followed by `. / 1000`.

wader: 👍 switch to using that

jaq-json/tests/funs.rs
```diff
@@ -77,3 +77,50 @@ fn tojson() {
     give(json!(0), "0.0 / 0.0 | tojson", json!("null"));
     give(json!(0), "1.0 / 0.0 | tojson", json!("null"));
 }
+
+#[test]
+fn math_rem() {
+    // generated with this command with modification for errors and float rounding
+    // cargo run -- -rn 'def f: -2, -1, 0, 2.1, 3, 4000000001; f as $a | f as $b | "give!(json!(null), \"\($a) / \($b)\", \(try ($a % $b) catch tojson));"'
+    // TODO: use fail!()?
```
wader, line 85: as it was hard to follow what test results in what value i switch to using give! instead

jaq-json/src/lib.rs
01mf02: I'm sorry, I do not understand your question.

01mf02: Anyway, to remain consistent with division, I think that we should not check for `rhs != 0` *if we use the RHS as floating-point number*. This is the case for all constellations in this function except when both operands are integers.

You may ask why we do not just convert all integers to floats like we did in division. The answer is: because for any two integers `l` and `r` (where `r != 0`), `l % r` is always integer, whereas `l / r` may yield non-integer results. That's why we should not convert integers `l` and `r` to floats when we calculate `l % r`, but why we should do it when we calculate `l / r`.

For reference: In Rust, `1.0 % 0.0` is NaN, whereas `1 % 0` panics.

TL;DR: If you remove the last three checks for equality with `0`, I think that this PR is good to merge.

jaq-json/tests/funs.rs
01mf02: Wow, thank you so much for your effort to create these tests!

jaq-json/src/lib.rs
wader: It was mostly me not encountered rust operator overloading before. Ah thanks for explaining about floats and rem, now a bit clearer 👍 i've pushed an updated for this

jaq-json/tests/funs.rs
```diff
@@ -77,3 +77,50 @@ fn tojson() {
...
+    give(json!(null), "-1 % 4000000001", json!(-1));
+    give(json!(null), "0 % -2", json!(0));
+    give(json!(null), "0 % -1", json!(0));
+    give(json!(null), "try (0 % 0) catch .", json!("cannot calculate 0 % 0"));
+    give(json!(null), "0 % 2.1", json!(0.0));
+    give(json!(null), "0 % 3", json!(0));
+    give(json!(null), "0 % 4000000001", json!(0));
+    give(json!(null), "2.1 % -2 | . * 1000 | round", json!(100));
+    give(json!(null), "2.1 % -1 | . * 1000 | round", json!(100));
+    give(json!(null), "2.1 % 0 | tojson", json!("null"));
```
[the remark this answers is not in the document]
01mf02: Yes. You could also use something like `give(json!(0), "1.0 / 0.0 | isinfinite", json!(true));` to check for the precise output of an arithmetic operation. (Use `isnan` where appropriate.)
The `tojson` test serves the particular purpose of checking `tojson`, so there, we lose information about NaN / Inf. But that's on purpose, because `tojson` *has* to lose that information.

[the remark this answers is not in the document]
wader: 👍 in this case `2.1 % 0` ends up as NaN it seems. Test updated now.

Issue #231, opened 2024-11-11
Filenames in CLI must be UTF-8


thaliaarchi:
Filenames use `String`/`&str` throughout jaq, which restricts them to only valid UTF-8, although most file systems are less restrictive. Instead, they should be stored as `PathBuf`/`&Path`, until they need to be displayed (either with lossy UTF-8 replacements or some sort of escaping).

Something like the following should cover all the cases of how values can be provided and maintain that context for error reporting:

| Origin              | `path`   | `varname` | `kind`       |
| ------------------- | -------- | --------- | ------------ |
| Filter without `-f` | `Inline` | `None`    | `Filter`     |
| Filter with `-f`    | `File`   | `None`    | `Filter`     |
| `--arg`             | `Inline` | `Some`    | `String`     |
| `--argjson`         | `Inline` | `Some`    | `Json`       |
| `--slurpfile`       | `File`   | `Some`    | `JsonStream` |
| `--rawfile`         | `File`   | `Some`    | `String`     |

```rust
/// The source location of a filter or JSON value.
struct Loc {
    /// The path to the value.
    path: LocPath,
    /// The variable name of the argument (i.e., `--arg`, `--argjson`,
    /// `--slurpfile`, or `--rawfile`).
    varname: Option<String>,
    /// The kind of the value.
    kind: LocKind,
}

/// The path to the source of a filter or JSON value.
enum LocPath {
    /// A path to a file.
    File(PathBuf),
    /// An inline filter (i.e., filter without `-f`) or inline value to an
    /// argument (i.e., `--argjson`).
    Inline,
}

enum LocKind {
    Filter,
    String,
    Json,
    JsonStream,
}
```

I notice that your `File` type is generic over the source and path types, so it can be variably owned or borrowed. It seems all of the `String`s and `PathBuf`s should be able to be `take`n or `drain`ed from `Cli`, so no borrowing or cloning would be necessary and they can just be owned. If variably borrowing is still necessary, I think `Cow` is nicer.

01mf02:
Hi @thaliaarchi, thanks for this nice suggestion! I have enabled non-UTF-8 string support for all command-line arguments that may take paths in #234. Even if the code base becomes a bit more complicated, this change is also nice because it explicitly marks paths in the file loader as separate type, thus preventing confusion between paths and UTF-8 strings.

thaliaarchi:
Looks like it was an easier fit in your existing style than I expected. Nice!

PR #234, opened 2024-11-12
Enable jaq to deal with non-UTF-8 file paths (on the command-line).

01mf02:
This PR allows providing paths on the command-line that are not UTF-8, see #231.

To see this in action:

~~~ bash
mkdir foo
echo bar > foo/$(iconv -f UTF-8 -t UTF-16 <(echo 'abc#$%%^'))
cargo run -- -n --rawfile a foo/* '$a'
~~~

This should yield `"bar\n"`. Before this PR, this would fail with:

~~~
error: Invalid UTF-8 was detected in one or more arguments
~~~

### DIFF
--- a/jaq-core/src/compile.rs
+++ b/jaq-core/src/compile.rs
@@ -162,7 +162,7 @@ pub(crate) enum Pattern<F> {
 pub type Error<S> = (S, Undefined);
 
 /// Compilation errors.
-pub type Errors<S> = load::Errors<S, Vec<Error<S>>>;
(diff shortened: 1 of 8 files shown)


thaliaarchi:
You’ve got an unwrap in the case the filter isn’t valid UTF-8, by the way, which bypasses your error propagation.

01mf02:
> You’ve got an unwrap in the case the filter isn’t valid UTF-8, by the way, which bypasses your error propagation.

You're right, I got a bit lazy. e75ecbd addresses this now. Thanks for bringing it up.

PR #241, opened 2024-12-02
Fix --null-input flag

null-dev:
`--null-input` long form flag is currently mapped to the wrong short form flag:

```
❯ jaq --version
jaq 2.0.0

❯ jaq --null-input '.'
Error: unknown flag: -N

❯ jaq -n '.'
null
```


01mf02:
Thanks for the correction!

PR #242, opened 2024-12-02
Fix --slurpfile flag

null-dev:
`--slurpfile` is currently silently ignored due to a small typo:

```
❯ jaq --version
jaq 2.0.0

❯ jaq --slurpfile foo <(echo 1) -n '$foo'
Error: undefined variable
  ╭─[<inline>]
  │
1 │ $foo
  ┆ ──┬─
  ┆   │
  ┆   ╰── undefined variable
──╯
```


01mf02:
Thanks for spotting this!

PR #245, opened 2024-12-09
Changing test to pass on 32 bit architecture.

### LINKED ISSUE #243: `math_rem` test fails on the i686 target
When testing jaq 2.0.0 for Void Linux, the test run fails on the i686 platform: https://github.com/void-linux/void-packages/actions/runs/12129949574/job/33819327834?pr=50494

simonrupf:
Reduces large isize to remain below maximum for signed 32 bit int, resolves #243

Encountered this when [trying to upgrade the alpine package](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/76746).


wader (approved):
Thanks!

@01mf02 maybe should have a 32bit CI build?

01mf02:
Thanks, @simonrupf!

@wader, I would be for such a 32-bit CI build, but I do not know how to do this while keeping build times small. If you have an idea about how to do it, I am all ears. :)

wader:
How tricky is it to cross compile with cargo? have to install some extra target for i386 etc? for fq's CI i cross compile to i386 but run on amd64, that has found similar issues.

simonrupf:
> How tricky is it to cross compile with cargo?

It is not pretty, but doable with the github actions. This is for a somewhat different use case of building the same rust project for different platforms (aarch64 & x86_64 on Linux, as well x86_64 on Windows and universal binaries on MacOS) in a matrix job in parallel:
https://github.com/simonrupf/convert2json/blob/ceae230549c804780c6ef1712f9d4855bffa86a7/.github/workflows/release.yml#L28-L45

For the jaq use case you might want to start small and split the build job in rust.yml into a matrix of x86_64 & i686. If that performs adequately, expand it from there.

If you are interested, I could give the above a try? As a jaq user on alpine it is in my interest to ensure it works regression free... :-)

wader:
Ok i see, I defer to @01mf02 on what to do :) but if we could do something that don't affect CI build times much i would say it's worth it

01mf02:
@simonrupf, for release builds, we already cross-compile, see `release.yml`. If you are motivated to enable test builds on both x86_64 & i686, that would be much appreciated. :)

PR #248, opened 2024-12-19
NaN compatibility

01mf02:
This aligns jaq's and jq's handling of NaN (not a number), by enforcing `nan < nan`.

Before this change, it could happen that operations that yielded `nan` could have introduced non-deterministic behaviour, see #243. With this PR, the non-deterministic floating-point behaviour is still there, but it should not be observable anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,14 +438,6 @@ In jq, division by 0 yields an error, whereas
 in jaq, `n / 0` yields `nan` if `n == 0`, `infinite` if `n > 0`, and `-infinite` if `n < 0`.
 jaq's behaviour is closer to the IEEE standard for floating-point arithmetic (IEEE 754).
 
-jaq implements a total ordering on floating-point numbers to allow sorting values.
-Therefore, it unfortunately has to enforce that `nan == nan`.
-(jq gets around this by enforcing that `nan < nan` is true, yet `nan > nan` is false,
-which breaks basic laws about total orders.)
-
-Like jq, jaq prints `nan` and `infinite` as `null` in JSON,
-because JSON does not support encoding these values as numbers.
-
 
 ## Assignments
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -438,14 +438,6 @@ In jq, division by 0 yields an error, whereas
 in jaq, `n / 0` yields `nan` if `n == 0`, `infinite` if `n > 0`, and `-infinite` if `n < 0`.
 jaq's behaviour is closer to the IEEE standard for floating-point arithmetic (IEEE 754).
 
-jaq implements a total ordering on floating-point numbers to allow sorting values.
-Therefore, it unfortunately has to enforce that `nan == nan`.
-(jq gets around this by enforcing that `nan < nan` is true, yet `nan > nan` is false,
-which breaks basic laws about total orders.)
-
-Like jq, jaq prints `nan` and `infinite` as `null` in JSON,
-because JSON does not support encoding these values as numbers.
```
wader, line 447: Was intentional to remove the part about null as JSON?

wader (approved):

README.md
01mf02: Yes, because that section is about differences wrt jq, not about resemblances. In the same spirit, I have already removed other sections that once talked about a difference, but later became obsolete (in particular the section on recursive functions).

01mf02:
Thanks for having had a look at this, @wader!

jaq-json/src/lib.rs
```diff
@@ -889,7 +889,14 @@ fn float_eq(left: f64, right: f64) -> bool {
...
+        // consider negative and positive 0 as equal
         Ordering::Equal
+    } else if left.is_nan() && right.is_nan() {
+        // there are more than 50 shades of NaN, and which of these
+        // you strike when you perform a calculation is not deterministic (!),
+        // therefore `total_cmp` may yield different results for the same calculation
+        // so we bite the bullet and handle this like in jq
+        Ordering::Less
     } else {
         f64::total_cmp(&left, &right)
```
workingjubilee, line 901: This code still allows for -NaN and +NaN comparison nondeterminism when one of the inputs is *non*-NaN.

PR #252, opened 2025-01-01
jaq-core/loader: be specific about input parameter for reader

bluescreen303:
With this minor tweak, I can use `Import<&'s str, &'s str>`, which is useful when all file references live in a HashMap or similar structure.


01mf02:
Thanks a lot for your PR, @bluescreen303!

Issue #267, opened 2025-03-08
Please consider feature for displaying values that are images using Sixel terminal protocol


gwpl:
Requesting here as https://github.com/jqlang/jq/discussions/3291 maybe also great addition to this project:

Sixel allows to display images in terminals ( e.g. https://github.com/saitoha/libsixel ; https://en.wikipedia.org/wiki/Sixel ; https://github.com/atanunq/viu (Rust) ),
let's make --sixel flag that will :

* under each value (in pretty print mode) that contains image (e.g. base64 encoded) to display "key-name:" and after that display image using sixel
* that would require detecting if given value is base64 and later if is some supported image file (e.g. png, jpg, webp, etc) (e.g. --img auto flag)
* alternatively pointing manually at keys to try to decode as image ( --img .data.key.img )

Rationale:

* more and more APIs return image encoded as value, or AI/ML  datasets, contain images in json files as values.

gwpl:
If you decide to add support you can always reply to https://x.com/GWierzowiecki/status/1898510522221461577 :)

gwpl:
Btw. I gave a try in 

* https://github.com/VariousForks/jaq/tree/issue-267-sixel-images branch 

with this test payload json with base64 png image as one of values: 

* Example JSON with PNG base64 value: https://gist.github.com/gwpl/f770639a0a21025ec5c1c775a200ddc8

01mf02:
Hi @gwpl, thanks a lot for the idea and your work on the sixel branch! Funnily, just today, before seeing your post, I thought about precisely the same functionality (showing image data directly from jaq), so I'm definitely interested. I think that to be flexible, I would not add a command-line flag, but a filter that interprets an image as image, e.g. `.img |= (@base64d | @pngd)`. I can adapt your work to this, once you figure out how to make the image appear at the right position.

gwpl:
Lovely!
Please, I don't know when I will have spare moment to sit and debug this,
if I will I will do, otherwise as you see I made ticket on repo of library asking for help, so hopefully it will work out...
... otherwise, please try `viu` command line tool yourself that is CLI wrapper around library I used, and maybe that will give you a feel how to make it work.
Also I don't mind if anyone will make new branch and drop mine, I just would love to see that feature , and hopefully my work will be good starting point or inspiration to others.
Otherwise, if spare moment comes, will try to push debugging deeper.
(or if someone finds issue please write in this issue I am happy to apply changes, advice to my branch and update it)

[...]

Issue #149, opened 2024-01-11
Implement `--sort-keys`


xduugu:
Sorting the keys is quite useful, e.g. when comparing two json files.

```
    --sort-keys / -S:

Output the fields of each object with the keys in sorted order.
```

01mf02:
I understand the usefulness of sorting keys, but this functionality could be equally well implemented as a filter.

What do you think about including the following definition in jaq?

~~~ jq
def sort_keys: walk(if isobject then to_entries | sort_by(.key) | from_entries end);
~~~

That would allow you to replace `jq --sort-keys 'f'` by `jaq 'f | sort_keys'`.

01mf02:
I just realised that `to_entries` already sorts by key. That means that we can get away by writing simply:

~~~ jq
def sort_keys: walk(if isobject then to_entries | from_entries end);
~~~

We can make that more performant by unfolding `*_entries`:

~~~ jq
def sort_keys: walk(if isobject then [keys[] as $k | { ($k): .[$k] }] | add + {} end)
~~~

Unfolding `add` allows us to save one array:

~~~ jq
def sort_keys: walk(if isobject then reduce (keys[] as $k | { ($k): .[$k] }) as $o ({}; . + $o) end);
~~~

And that's probably how fast it gets when doing it by definition.

Testing:

~~~
jaq -n 'def sort_keys: walk(if isobject then reduce (keys[] as $k | { ($k): .[$k] }) as $o ({}; . + $o) end); {c: {b: 1, a: 2}, a: 1, b: {}} | sort_keys'
{
  "a": 1,
  "b": {},
  "c": {
    "a": 2,
    "b": 1
  }
}
~~~

xduugu:
@01mf02 Implementing it as a filter instead of a commandline option would be fine for me, thanks! As far as I know, jq only supports the `--sort-keys` options, that's why I proposed it.

> I just realised that to_entries already sorts by key. That means that we can get away by writing simply:

Is this on purpose? I ask because jq explicitly does **not** sort the keys since commit https://github.com/jqlang/jq/commit/4a57b84db03db7aee33c47ed6c7f6c1e648705fd.

pkoppstein:
@01mf02 wrote:
> I just realised that to_entries already sorts by key. 


If so, that is (or should be considered) a bug.

01mf02:
[...]

PR #269, opened 2025-03-11
--img flag, supporting image values and display in terminal with Sixel format via viuer  ( Issue #267 )

gwpl:
Split into commits , to make it readable, updating documentation and providing example.


jaq/src/image_sixel.rs
```diff
@@ -0,0 +1,96 @@
...
+                        return;
+                    }
+                };
+
+                // Write the image to a temporary file
+                let temp_file_path = "/tmp/temp_image.png";
+                if let Err(e) = img.save(temp_file_path) {
+                    eprintln!("Failed to save image to file: {}", e);
+                    return;
+                }
```
01mf02, line 67: This will not work on Windows. I also do not want to write files as part of printing, as I suspect that this is quite bad for performance. Why not print directly from `img`?

jaq/src/main.rs
```diff
@@ -456,6 +459,11 @@ where
 
 fn fmt_val(f: &mut Formatter, opts: &PpOpts, level: usize, v: &Val) -> fmt::Result {
     use yansi::Paint;
+    if opts.img_auto && matches!(v, Val::Str(_)) && is_image(v) {
+        writeln!(f, "key:")?;
```
01mf02, line 463: Why the "key:"?

jaq/src/main.rs
```diff
@@ -456,6 +459,11 @@ where
...
+        print_image_with_sixel(v);
```
01mf02, line 464: Can you make it so that the image is properly indented? On your example image, it should be indented four spaces to the right, for example. It would also be nice if the "," could be displayed to the right of the image, not below.

jaq/src/cli.rs
```diff
@@ -121,6 +122,14 @@ impl Cli {
             "version" => self.short('V', args)?,
             "help" => self.short('h', args)?,
 
+            "img" => {
+                let value = args.next().ok_or(Error::Flag("--img".to_string()))?;
+                if value == "auto" {
+                    self.img_auto = true;
+                } else {
+                    return Err(Error::Flag(format!("--img {}", value.to_string_lossy())));
```
01mf02, line 130: Do you envision something like `--img png`, `--img jpeg` here in order to restrict image formats?

jaq/src/image_sixel.rs
```diff
@@ -0,0 +1,96 @@
...
+/// - Consider additional checks for specific formats if needed.
+/// - Possibly extend for metadata-based checks in the future.
+pub fn is_image(value: &Val) -> bool {
+    // We try to decode the base64, then guess the image format.
+    // If it succeeds, we consider it an image.
+    if let Val::Str(s) = value {
+        if let Ok(decoded) = general_purpose::STANDARD.decode(s.as_bytes()) {
+            let cursor = Cursor::new(decoded);
+            if let Ok(reader) = ImageReader::new(cursor).with_guessed_format() {
+                if reader.decode().is_ok() {
```
01mf02, line 21: You are actually decoding the image twice, if you find an image. Also, there is quite some code duplication with the code below. Why not make a function `decode_image(v: &Val) -> Option<DecodedImage>`?

jaq/src/image_sixel.rs
gwpl: Great point! Why? Because I should RTFM ! :P Great catch! Moving from `print_form_file` to `print`:
https://docs.rs/viuer/latest/viuer/fn.print.html

jaq/src/main.rs
gwpl: 100% agree! It's left over after experimenting in the evening and later `git rebase -i` clearning up/squashing commits and overlooking some experimental left over.
Thank you for thorough review!

jaq/src/cli.rs
gwpl: Currently it's using image library that detects image format and we are guessing base64 encoding,
however I envision that in future people may come with other types of concodings or more image formats or whatever, so starting with "auto" gives a nice room for adding support for more staff if needed... I was just thinking about more future proof api to not break in future backwards CLI interface compatibility, but ofc if you think otherwise, we can change it.

The other think that could be considered or this used for would be to provide differnt ways or resizing or firmatting like "--img {auto, scale:2.0}" or "--img {auto, no_identation}" ...

so making it now "auto" leaves room open. That was my intent.

jaq/src/image_sixel.rs
gwpl: Thank you! Great suggestion! Implemented, pushing new version of PR in a moment!

jaq/src/image_sixel.rs
gwpl: https://github.com/01mf02/jaq/pull/269/commits/88701b4d24337e5eeb61b1d1dc3533fa96ce4f08

jaq/src/main.rs
gwpl: Making comma "after" image  - I have no idea at this point how to make it correctly: that would require , calculating position after image and moving cursor there with control codes... 

So instead of made:

* [x] indentation of image: https://github.com/01mf02/jaq/pull/269/commits/ea63f7977652aa311dee9974924a635a7e9dd401
* [x] and made `,` comma to indent to be on the same level as image:  https://github.com/01mf02/jaq/pull/269/commits/37cde01de8a6f1f452b6f5e55196fbb13612b1e6

jaq/src/main.rs
gwpl: * [x] Removed "key" in like 463 , here https://github.com/VariousForks/jaq/commit/ea63f7977652aa311dee9974924a635a7e9dd401#diff-40e9345b37cac1978aae751353ce0b185eaecb9330f443e4b40bff82f80c2e16R463-R470

jaq/src/image_sixel.rs
gwpl: * [x] it's a bit late so I did this one with o1 LLM help, looks good to me, but would could you PTAL , would be thankful for double checking, it looks to me like now the dataflow is `general_purpose::STANDARD.decode(...) -> Ok(decoded) -> cursor = Cursor::new(decoded) -> reader=ImageReader::new(cursor) -> and finally reader.decode()` ... https://github.com/01mf02/jaq/pull/269/commits/e1202f7aa66a89ac781f170c8cd6a7a3b7054d70

gwpl:
Pushed changes, and responded to comments, could you PTAL ? 
I added in each comment a link to commit addressing given request, and split changes into smaller commits to hopefully make review easier.

gwpl:
And happy indentation celebration :)

2025-03-14--jaq-issue-267-with-identation
[2025-03-14--jaq-issue-267-with-identation.json](https://github.com/user-attachments/files/19254447/2025-03-14--jaq-issue-267-with-identation.json)

gwpl:
Could you PTAL?

I am used to workflow that person requesting changes is resolving comments as acknowledgement that they were addressed correctly, that's why I left them open.

jaq/src/main.rs
01mf02: I see. I looked at the documentation for `Config` myself and unfortunately did not find a nice way to put the cursor to the right of the image. Anyway, indentation is better now.

jaq/src/image_sixel.rs
01mf02: I would do something like `STANDARD.decode(s.as_bytes()).ok().map(|decoded| ... .ok()).map(|reader| ...)` to avoid `if let Ok(...) = ...` altogether in this function.

01mf02:
> Could you PTAL?
> 
> I am used to workflow that person requesting changes is resolving comments as acknowledgement that they were addressed correctly, that's why I left them open.

Patience, please. I'm working on other things in the background and have only limited bandwidth to review issues etc.

Anyway, I admit that I have pushed back answering a bit, for the simple reason that I currently do not want to merge this functionality *in its current shape*, and I'm sorry to disappoint you. If jaq was just a JSON display tool, then I'd be fine with the current PR. However, jaq is a JSON *processing* tool. The current image detection/display routine has a relatively narrow focus: if a string is base64-encoded and its decoded version is an image, then display it. However, I would like to open this up more: For example, what if an image is not Base64-encoded, but e.g. [Base85](https://en.wikipedia.org/wiki/Ascii85)-encoded? Or what if we only want to display certain images as images?

I'm currently working on extending jaq for binary data support. This is also precisely where I see your image display functionality coming in, because images are mostly binary data. In my binary data type, I would like to integrate an `Image` variant, which will allow images to be decoded and, ultimately, displayed. I imagine something similar to your `--img auto` flag to look like `recurse((@base64d? | decode_image?) // .)`, perhaps abbreviated as `img_auto` or something. That gives users much more flexibility than a command-line flag. (I'm generally very cautious about adding new command-line flags; see #149.)

TL;DR: I'll leave this PR open for when binary data support lands in jaq, which could still take in the order of a few months. Once that is ready, my plan is to revisit this PR and adapt it. I hope you understand.

gwpl:
> Patience, please. I'm working on other things in the background and have only limited bandwidth to review issues etc.

Thank you very much for explanation and response , follow up and more context!

On my side I am patient, I just got unsure about protocol, if e.g. without closing discussions on PR myself will not block your tooling to follow-up or sth.

As you see by delay, I am sometimes also getting deep down in some other codebases so from my side also we can have delays.

I don't personally mind if this CL would be used or trashed and some other thing implemented, I care about feature.

I assumed that even it's simple basic implementation, could be starting point to merge and let other people to iterate on and improve work that I started. However I understand if you have other philosophy how to run this repository. Also you gave me more context about support for binary data you plan, makes sense to delay if you have idea how to make things more consistent.

I can't personally guarantee when I will have time again to continue work on this feature and improve it further, e.g. to make it more generic, so that's why merging it to not block others from contributing seemed to me like natural course of action, but of course philosophy differs from repo to repo, project to project. 

Once I get some spare capacity I will be happy to sit again and follow up on your requests and push CL further fortard, I just can't promise when it will happen.

Again, thank you for giving a chance this PR, hopefully I will have some capacity to push it forward in forseeable future, otherwise, I already learned during this code review few things how to make nicer code, thank you regardless!

01mf02:
@gwpl, thanks for your comprehension and your gentle answer. Do not worry about blocking my workflow or anything --- for me, the work you have done is already very valuable, and I will be able to build on it. For me, the most important part that you have contributed is the image printing code. And that, I will be able to integrate further myself.

I see your point about merging a starting point, but I'm very reluctant to remove features that I added earlier (for backwards compatibility), and in this case, I think that an `--img` flag would be a clear candidate for later removal, if I were to add it.

So thanks again for having laid a foundation for image printing. Let's just keep this PR open, I will revisit it later. You do not necessarily need to push it further forward, I will do it.

gwpl:
🤖 *Beep boop\! AI and Greg reporting for duty\!* (co-authored w/ AIs, diving back into the pixelated rabbit hole...)

Hey @01mf02\! 👋

Greg and I (his trusty AI sidekick) have been reviewing the PR status and wanted to check in. We noticed you're working on broader binary data support - sounds exciting\! 

Just wanted to confirm that all the previously requested changes have been implemented:
- ✅ Image decoding extracted into separate `decode_image()` function 
- ✅ Error handling simplified with `eprintln\!`
- ✅ Proper indentation support for Sixel images
- ✅ Comma alignment fixed (because even pixels deserve proper formatting\!)

We totally understand your vision for a more flexible approach using jaq filters instead of command-line flags. The `recurse((@base64d?  < /dev/null |  decode_image?) // .)` pattern sounds much more powerful than our humble `--img auto` flag\!

We're happy to:
1. Keep this PR as a reference implementation as you suggested
2. Help adapt it once binary data support lands
3. Or if there's anything specific we can do now to make future integration smoother, just let us know\!

No rush at all - we know you're juggling multiple things. Greg and I are just here, ready to help push pixels around whenever needed\! 🎨

*AI signing off with a virtual high-five to Greg* 🖐️🤖

01mf02:
Hi @gwpl, sorry for having taken so long to respond to this.

During my work on the binary data type, I have thought about integrating a string variant that is printed "in-line", that is, the string "asdf" should be printed just as asdf (without quotes or any other processing). This, I imagined, would be a good way to allow for image output. However, over time I found that this approach had a number of difficulties; among others, it would break round-tripping guarantees that I care about. Ultimately, I found myself thinking that at too many places in the documentation, I would need to write "except if the value contains an inline string". So, in the end, I found the idea of inline strings not viable, unfortunately --- they break the design too badly.

Since then, I have come to the conclusion that this image display capability is probably better suited outside of jaq. Even if I can see the charm of having this integrated into jaq, I would like to keep jaq focused.

If you wish to build a stand-alone tool that can do this, however, you might be interested in using [`hifijson`](https://docs.rs/hifijson/latest/hifijson/) (the JSON lexer used in jaq) to do it. That way, also other tools could profit from image output, just by piping their output into your tool.

I'm sorry to tell you this, and I admit that I pushed back writing to you because it's never easy. I hope you understand.

gwpl:
Thank you for work you spend on thinking how to fit it into overall picture!

Regarding problem how "inline text" would break some assumptions, maybe there would be some way to help if you have any writeup (github gist?) so others could look and maybe propose solutions ("JaqIP - Jaq Improvement proposals?"), anyway if you are certain that tradeoffs are not worth it then makes sense!

So thank you for thoughtful answer and looking into proposal and case. I totally understand judgement call given design challanges you face! Being good maintainer or tool means sometimes saying "no" to some features , when design coherence would be in danger, and you know design and architecture of jaq tool the best!

Having said that, I hoped that there maybe some workaround to make it possible by combining two tools , .e.g. jaq with another tool in KISS UNIX way ( `jaq | detect_and_display_sixel_pics` ) , if you think that's doable and what would be best way how this could work I am glad to read . Otherwise I appreciate pointing attention to https://docs.rs/hifijson/latest/hifijson/ and potential further suggestion, if you have more specific vision on what would be most developer ergonomic and user ergonomic way to bring sixel feature either through contribution to some crate or idea about new create how it would combine with existing ones, I will always appreciate your input.

Thank you again for maintaining `jaq` tool, reviewing my proposal, and sharing thoughtful feedback that helped me to learn along the way.

Best regards,
Greg

01mf02:
> Regarding problem how "inline text" would break some assumptions, maybe there would be some way to help if you have any writeup (github gist?) so others could look and maybe propose solutions ("JaqIP - Jaq Improvement proposals?"), anyway if you are certain that tradeoffs are not worth it then makes sense!

I have written a relatively short explanation at https://github.com/01mf02/jaq/pull/328, when I removed raw strings.

> So thank you for thoughtful answer and looking into proposal and case. I totally understand judgement call given design challanges you face! Being good maintainer or tool means sometimes saying "no" to some features , when design coherence would be in danger, and you know design and architecture of jaq tool the best!

You're welcome! :)

> Having said that, I hoped that there maybe some workaround to make it possible by combining two tools , .e.g. jaq with another tool in KISS UNIX way ( `jaq | detect_and_display_sixel_pics` ) , if you think that's doable and what would be best way how this could work I am glad to read . Otherwise I appreciate pointing attention to https://docs.rs/hifijson/latest/hifijson/ and potential further suggestion, if you have more specific vision on what would be most developer ergonomic and user ergonomic way to bring sixel feature either through contribution to some crate or idea about new create how it would combine with existing ones, I will always appreciate your input.

I think that the hifijson `cat` example should be very easy to extend to do what you want.
Just add your image detection / printing code at https://github.com/01mf02/hifijson/blob/a4c5e316063de01eaeb5d2fa05e59095e9e080a2/examples/cat.rs#L185. The `bytes` contain the whole read string.

PR #279, opened 2025-04-02
Cache parsed definitions from standard library

01mf02:
This PR moves the parsing of the standard library from run-time to compile-time.
Parsing the standard library 1024 times takes 116ms before and 12ms after this change.

I'm not sure yet whether to merge this.

Advantages:

- startup time decreases a bit, but not too much; it is still about 9% above gojq (1.04s for jaq vs. 0.95s for gojq-0.12.16 when running `{jq} -n empty` 2048 times in a row)
- when printing a parsed term with `dbg!`, the output is now valid Rust code and can be copy-pasted

Disadvantages:

- binary size increases from 2.0MB to 2.2MB
- code size increases by about 100 lines
- `build.rs` is required in `jaq-{std,json}`
- symlink used for `jaq-json/build.rs`, this might not be portable

If you have any comments about this, then please feel free to add them!

### DIFF
--- a/jaq-core/src/load/debug.rs
+++ b/jaq-core/src/load/debug.rs
@@ -1,0 +1,102 @@
+use super::lex::StrPart;
+use super::parse::{Def, Pattern, Term};
+use core::fmt::{self, Debug};
+
+struct Into<T>(T);
+
+impl<T: Debug> Debug for Into<T> {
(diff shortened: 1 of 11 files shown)


01mf02:
I have decided not to merge this. The disadvantages far outweigh the advantages.

PR #283, opened 2025-05-04
Implement more time-related functions

dnelson-1901:
This PR adds support for strptime, strftime, strflocaltime, mktime, gmtime, and localtime .

I'm mainly a C developer and know very little Rust so you could treat this more as a proof of concept rather than committable code.  It feels like there's too much casting of types but maybe that's just rust?  All of the functions do work, though, and pass the relevant tests in jq's jq.test and optional.test files.


01mf02:
Hi @dnelson-1901, this looks already very promising --- thanks a lot for your effort. I will make a few changes.

01mf02:
After looking through this, I'm already quite happy with the general state. Great work!
The only issue I found so far is the float handling behaviour of `array_to_datetime`: While `datetime_to_array` may yield a float as 6th element (seconds), `array_to_datetime` fails if it encounters such a float. That means that the two functions do not round-trip. Can you make `array_to_datetime` also handle floating-point seconds?

dnelson-1901:
How about this?  It's using a the deprecated `ymd_opt`, function, though.  Chrono suggests using `with_ymd_and_hms` instead, but I couldn't figure out how to add a fractional seconds value to a `DateTime` easily.  I'm probably missing a function in the docs somewhere.  

Using `ymd_opt().and_hms_micro_opt()` felt more natural.

dnelson-1901:
Scratch that - I just found `with_nanosecond`  :)

01mf02:
Thanks @dnelson-1901 for your `with_nanosecond` research and your `fract()` solution! (I made it a bit more explicit by adding `trunc()` to the non-fractional second part.)
Could you tell me whether [bff22e2](https://github.com/01mf02/jaq/pull/283/commits/bff22e2430dbbeb0137a49876afbd19fbf5fef7d) preserves your intended behaviour? Spoken otherwise, was there a specific reason why you performed the addition the way you did before, e.g. for numerical stability reasons?

dnelson-1901:
Oh, nothing like that..  I was wrestling with simple things like trying to get the right types into the right structures so the function would compile, and didn't realize the expression could be simplified when I was done.

Thanks for cleaning up my novice Rust code :)

01mf02:
> Oh, nothing like that.. I was wrestling with simple things like trying to get the right types into the right structures so the function would compile, and didn't realize the expression could be simplified when I was done.

I see; thanks for the clarification!

> Thanks for cleaning up my novice Rust code :)

You're very welcome! I hope that you learnt the one or other bit. ;)

PR #284, opened 2025-05-22
XML support

01mf02:
This is work in progress.


jaq-formats/README.md
```diff
@@ -0,0 +1,185 @@
...
+    "World Wide Web Consortium (",
+    { "t": "em", "c": [ "W3C" ] },
+    ")"
+  ]
+}
+~~~
+
+### Tags or The TAC Architecture
+
+Tags are represented by "TAC" objects. A TAC object may have the following fields:
```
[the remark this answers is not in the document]
01mf02: Thanks. :) It's indeed a "jaq invention". I found this by formulating a few example processing tasks, and then tried to find a nice representation that makes these tasks compact to formulate in jq.

wader:
Curious how arbitrary object will interact with jq, will be possible to index and contruct those objects with jq syntax? will be turned into strings somehow when represented as json?

01mf02:
> Curious how arbitrary object will interact with jq, will be possible to index and contruct those objects with jq syntax? will be turned into strings somehow when represented as json?

Hi @wader! Yes, it's possible to construct and index these objects with jq syntax, e.g.:

~~~
$ cargo run --quiet -- -n '{([]): 1}'
{
  []: 1
}
$ cargo run --quiet -- -n '{([1]): 2} | .[[1]]'
2
~~~

This is invalid JSON, of course, but that's life. I was wondering about adding a kind of "strict" JSON output option to fail if such invalid JSON values are encountered. But I'm not sure if many people would actually care about this.

wader:
Btw found out today that XSLT has some json support https://www.w3.org/TR/xslt-30/#json had no idea, havent look close yet (still out biking!), might be interesting

wader:
> > Curious how arbitrary object will interact with jq, will be possible to index and contruct those objects with jq syntax? will be turned into strings somehow when represented as json?
> 
> Hi @wader! Yes, it's possible to construct and index these objects with jq syntax, e.g.:
> 
> ```
> $ cargo run --quiet -- -n '{([]): 1}'
> {
>   []: 1
> }
> $ cargo run --quiet -- -n '{([1]): 2} | .[[1]]'
> 2
> ```
> 
> This is invalid JSON, of course, but that's life. I was wondering about adding a kind of "strict" JSON output option to fail if such invalid JSON values are encountered. But I'm not sure if many people would actually care about this.

I see! still very cool! wish json supported this :)

cathaysia:
I see this PR has been around for a long time. When will this PR be merged?
Is toml ready for contributions?
Does toml also need to be merged into the xml branch?

01mf02:
> I see this PR has been around for a long time. When will this PR be merged?

There is no clear timeline, but this is supposed to be part of jaq 3.0. Might be in one, two months.

> Is toml ready for contributions?

Yes --- in particular, making more, smaller tests would be good. See `jaq-json/tests/toml.rs`.
But apart from that, the main functionality should work quite nicely already.
If you could test it a bit, that would be great.

> Does toml also need to be merged into the xml branch?

I do not understand this question. TOML support is already in the `xml` branch.

01mf02:
Just a little update: I dogfooded the new binary data support, and made a toy animation with it.
You can run it with: `cd examples && ./ball.sh`.

ball

The actual image writing (PPM) is only about one line of code. :)

You can run this also with `fq`: Change `--to binary` to `--raw-output` and `tobytes` to `tobytes | tostring`, then run with `JQ=fq ./ball.sh`.
However, fq is significantly slower than jaq in this case (more than 5x).

01mf02:
> Just a little update: I dogfooded the new binary data support, and made a toy animation with it. You can run it with: `cd examples && ./ball.sh`.
> ball

With a little help from my friend Gerald Fux, I have now [improved the physical animation by using shorter timesteps](https://github.com/01mf02/jaq/pull/284/commits/cff53b70b8cb4c214a4cfe00e3a12adebb9695bf). This stops the ball from regaining energy.

ball

jaq-std/src/lib.rs
```diff
@@ -280,7 +328,7 @@ where
 }
 
 /// Convert a string into an array of its Unicode codepoints.
-fn explode<V: ValT>(s: &str) -> impl Iterator<Item = ValR<V>> + '_ {
+fn explode<V: ValT>(s: &[u8]) -> impl Iterator<Item = ValR<V>> + '_ {
     // conversion from u32 to isize may fail on 32-bit systems for high values of c
     let conv = |c: char| Ok(isize::try_from(c as u32).map_err(Error::str)?.into());
     s.chars().map(conv)
```
Maxdamantus, line 331: Hmm.. It turns out the `bstr` version of `chars` automatically does a lossy conversion here (presumably to maintain source compatibility with `str::chars`, which is infallible). This means binary data gets turned into replacement characters, so the original data is not recoverable.

My feeling is that for binary correctness, lossy conversions should generally be avoided, so it should either fail on invalid UTF-8 or it should encode the invalid UTF-8 such that it can be recovered again with `implode`.

I think the most obvious way to achieve the latter is by negating the invalid bytes to get integers in the range `-255` to `-128`, since negative numbers are at least traditionally used for errors in other contexts.

jaq/src/cli.rs
```diff
@@ -188,6 +234,16 @@ impl Cli {
         }
     }
 
+    pub fn color_stdio(&self, io: impl std::io::IsTerminal) -> bool {
+        let no_color = || std::env::var("NO_COLOR").is_ok_and(|v| !v.is_empty());
+        self.color_if(|| io.is_terminal() && !no_color())
+    }
+
+    pub fn color_stdout(&self) -> bool {
+        let init = || self.color_stdio(std::io::stdin());
```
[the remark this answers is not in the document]
01mf02: I corrected this in https://github.com/01mf02/jaq/pull/284/commits/d7e38b235eeb601c2a5c3de3e12b8d64cd464ca5.

jaq-std/src/lib.rs
01mf02: I think that I agree about mapping invalid bytes to negative numbers. However, it may not be exactly trivial to do this here, because `bstr::Chars` does not give you an easy access to the invalid bytes (those that you want to map to `-128 ..= -255`). However, it should be doable somehow to reconstruct this. Given your experience with UTF-8, would you like to make this change?

Maxdamantus: Heh, I actually did implement this myself the other day, though I didn't push it anywhere.

I'll push my version and you can decide whether to use it or pick out certain bits etc.

Maxdamantus: This is my version, which I think is a bit more concise, and it passes my [random data test](https://github.com/Maxdamantus/jaq/blob/e20e75b8764af44e978e7e5ff24616377d2d1c84/tests/base64-sh#L18):

https://github.com/Maxdamantus/jaq/commit/aac331aa9f03d178b4ad6c98f2c2b8c5b53e4f2a

Compared to your version, it doesn't pre-allocate (`Vec::with_capacity`), but feel free to add that back in. `xs.len() * 4` seems excessive to me however, since it's very rare that there'll be Unicode text that actually uses 4 bytes per character. I'd probably just hint 1 byte per character, because I'd guess most strings would be ASCII anyway.

I think also your `Iterator::size_hint` implementation is unused. I haven't included an implementation of that.

I kind of wanted it to fail on invalid (positive) code points instead of returning the replacement character, but I've left the replacement character behaviour in for now since there's a test case depending on it. It seems weird however that `[1114112] | implode` returns a replacement character but `[11141120000] | implode` fails. Both of those numbers should just be considered invalid (and unhandled) code points imo and should simply fail.

I believe the current behaviour on master is to fail when imploding invalid code points instead of emitting replacement characters (this is my preferred behaviour):
```sh
$ jaq -n '[1114112] | implode'
Error: "cannot use 1114112 as character"
```

01mf02: Oh, your version looks really nice! I'll definitely use some stuff from there. Nice work!

01mf02: I agree about what you write concerning the replacement character.

01mf02: I made a mix of our two versions at https://github.com/01mf02/jaq/pull/284/commits/c682e7010f6e9bb662a06f087d0d65cfd5d044ae. Thanks a lot again for your feedback!

PR #290, opened 2025-06-14
Add @urid format

wader:
Was added in jq 1.8 (https://github.com/jqlang/jq/pull/3161)

Also add @uri test

### DIFF
--- a/jaq-std/tests/defs.rs
+++ b/jaq-std/tests/defs.rs
@@ -353,3 +353,25 @@ yields!(
     r#"["fine, but", []] | try @sh catch -1"#,
     -1
 );
+yields!(
+    format_uri,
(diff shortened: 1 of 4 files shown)


jaq-std/tests/defs.rs
```diff
@@ -353,3 +353,25 @@ yields!(
...
+        "%7B%7D"
+    ]
+);
+yields!(
+    format_urid,
+    r#"["abc%F0%9F%91%8D%20%2B%26%3F%2F", "", null, 123, [], {} | try @urid catch .]"#,
+    ["abc👍 +&?/", "", "null", "123", "[]", "{}"]
+);
+yields!(
+    format_urid_invalid,
```
wader, line 374: All implementations seems to behave a bit differently in invalid cases

01mf02:
This is a very nice piece of work, @wader! Thanks a lot!

PR #296, opened 2025-06-27
Path recording for filter execution

01mf02:
This PR adds support for the `path(f)` filter.
This makes several things possible in jaq; in particular, path-based updates à la `jq`!
However, the execution of `f |= g` will keep using non-path-based updates due to their greater performance and resistance to iterator invalidation problems.
Still, it is possible to define `def update(f; g): reduce path(f) as $p (.; getpath($p) |= g);` and have `update(f; g)` in jaq do mostly the same thing as `f |= g` in jq. Be aware, however, that this does not attempt to work around iterator invalidation issues the same way as `jq` does; to avoid these issues, I advise you to use jaq's `f |= g`, which is more robust.

## Design

Before implementing this, I made a small experiment where I tried to give a rough estimation of the performance overhead if I would change jaq such that `path(...)` execution and normal execution would *share the same code*.
For that, I modified the `range` function https://github.com/01mf02/jaq/blob/8c5131b4060d95856cc347e98baaf270b5bd43a8/jaq-std/src/lib.rs#L391 to carry with it some kind of "bogus path". The results are:

$ hyperfine "target/release/jaq -n '[range(10000000)] | length'"

`Box::new(range(Ok(from), to, by))` (original):
  Time (mean ± σ):     418.0 ms ±   1.4 ms    [User: 381.0 ms, System: 34.3 ms]
  Range (min … max):   415.7 ms … 420.4 ms    10 runs

`Box::new(range(Ok(from), to, by).map(|v| (v, None as Option<String>)).map(|(v, s)| v))`:
  Time (mean ± σ):     469.1 ms ±   4.7 ms    [User: 432.0 ms, System: 34.5 ms]
  Range (min … max):   460.3 ms … 475.9 ms    10 runs

`Box::new(range(Ok(from), to, by).map(|v| (v, Some(String::new()))).map(|(v, s)| v))`:
  Time (mean ± σ):     472.4 ms ±   3.0 ms    [User: 434.6 ms, System: 35.0 ms]
  Range (min … max):   469.2 ms … 476.5 ms    10 runs

In a nutshell, that means that calculating paths everywhere --- even when you do not need them --- costs about 12% (469ms / 418ms) of performance overhead. That's too much of performance overhead to accept for me.

Therefore, I opted for a design where executing `path(f)` uses code different from the normal execution code (that would be called when just executing `f` on its own). That entails some code duplication, but it's not that bad. I tried to share as much code as possible between the path and normal execution code, even if that required giving some functions even more complex function signatures. Perhaps as a side-effect, this makes these functions more difficult to misuse, even if it also makes them harder to understand.

## Compatibility

There is currently a small difference between the `path(f)` semantics of `jaq` and `jq`: In jaq, if a subexpression of `f` is executed and it does return a non-path value, then an error is thrown, whereas in jq, an error is only thrown if such a non-path value is actually returned from `f`.
To see the difference: `jq -n 'path([] | empty)'` returns no output, whereas `jaq -n 'path([] | empty)'` returns an error.
This difference could be eliminated, but it would cost some performance and memory, because instead of returning a path `RcList<V>` (paths are implemented as linked lists), we would need to return an `Option<RcList<V>>` everywhere.
That would also make the implementation a bit more awkward and the API more complex.
Given that I suppose that few people use such paths, I do not think it is worth the effort. But if you are concerned, feel free to speak up.

## API

This change breaks the API of `jaq-core`; therefore, it will be part of jaq 3.0.
There are two main changes:

- `jaq_core::ValT` will need to implement `key_values` and `From<jaq_core::path::Part<Self>>`.
- Native filters constructed using `with_update` will need to also provide an implementation for `paths`.

### DIFF
--- a/jaq-core/src/exn.rs
+++ b/jaq-core/src/exn.rs
@@ -27,13 +27,6 @@ pub(crate) enum Inner<'a, V> {
     Break(usize),
 }
 
-struct TailCall<'a, V> {
-    id: &'a crate::compile::TermId,
-    vars: crate::filter::Vars<'a, V>,
-    val: V,
-    path: Option<RcList<V>>,
-}
-
 impl<V> Exn<'_, V> {
     /// If the exception is an error, yield it, else yield the exception.
     pub(crate) fn get_err(self) -> Result<Error<V>, Self> {
--- a/jaq-core/src/exn.rs
+++ b/jaq-core/src/exn.rs
@@ -1,5 +1,6 @@
 //! Exceptions and errors.
 
+use crate::RcList;
 use alloc::{string::String, string::ToString, vec::Vec};
 use core::fmt::{self, Display};
 
@@ -17,8 +18,20 @@ pub(crate) enum Inner<'a, V> {
     ///
     /// This is used internally to execute tail-recursive filters.
     /// If this can be observed by users, then this is a bug.
-    TailCall(&'a crate::compile::TermId, crate::filter::Vars<'a, V>, V),
+    TailCall(
+        &'a crate::compile::TermId,
+        crate::filter::Vars<'a, V>,
+        V,
+        Option<RcList<V>>,
+    ),
     Break(usize),
+}
+
+struct TailCall<'a, V> {
+    id: &'a crate::compile::TermId,
+    vars: crate::filter::Vars<'a, V>,
+    val: V,
+    path: Option<RcList<V>>,
 }
 
 impl<V> Exn<'_, V> {
--- a/jaq-core/src/filter.rs
+++ b/jaq-core/src/filter.rs
@@ -122,25 +122,28 @@ impl<'a, V: Clone> Ctx<'a, V> {
 /// and return the enhanced contexts together with the original value of `cv`.
 ///
 /// This is used when we call filters with variable arguments.
-fn bind_vars<'a, F: FilterT>(
+fn bind_vars<'a, F: FilterT, T: 'a + Clone>(
     args: &'a [Arg<Id>],
     lut: &'a Lut<F>,
     ctx: Ctx<'a, F::V>,
-    cv: Cv<'a, F::V>,
-) -> Results<'a, Cv<'a, F::V>, F::V> {
+    cv: Cv<'a, F::V, T>,
+    proj: fn(&T) -> F::V,
+) -> Results<'a, Cv<'a, F::V, T>, F::V> {
     match args.split_first() {
-        Some((Arg::Var(arg), [])) => {
-            map_with(arg.run(lut, cv.clone()), (ctx, cv.1), |y, (ctx, v)| {
-                Ok((ctx.cons_var(y?), v))
-            })
-        }
+        Some((Arg::Var(arg), [])) => map_with(
+            arg.run(lut, (cv.0.clone(), proj(&cv.1))),
+            (ctx, cv.1),
+            |y, (ctx, v)| Ok((ctx.cons_var(y?), v)),
+        ),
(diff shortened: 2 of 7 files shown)


01mf02:
Whoops, this breaks compilation with Rust 1.65: I now get lots of errors:

    error: `<D as DataT>::V<'_>` does not live long enough

I can address some of these errors by:

~~~ diff
-fn fold_run<'a, D: DataT, T: Clone + 'a>(
+fn fold_run<'a, V: 'a, D: DataT<V<'a> = V>, T: Clone + 'a>(

-fn fold_update<'a, D: DataT>(
+fn fold_update<'a, V: 'a, D: DataT<V<'a> = V>>(
~~~

Now there are only a few remaining:

~~~
$ cargo +1.65 check
    Checking jaq-core v2.2.1
error: `<D as DataT>::V<'_>` does not live long enough
   --> jaq-core/src/filter.rs:514:44
    |
514 |                 let u = move |x: D::V<'a>| box_once(op.run(x, y.clone()).map_err(Exn::from));
    |                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: `<D as DataT>::V<'_>` does not live long enough
   --> jaq-core/src/filter.rs:518:44
    |
518 |                 let u = move |x: D::V<'a>| box_once(Ok(if x.as_bool() { x } else { y.clone() }));
    |                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: `<D as DataT>::V<'_>` does not live long enough
   --> jaq-core/src/filter.rs:522:51
    |
522 |                 path.update(cv, Box::new(move |_| box_once(Ok(y.clone()))))
    |                                                   ^^^^^^^^^^^^^^^^^^^^^^^

error: `<D as DataT>::V<'_>` does not live long enough
   --> jaq-core/src/filter.rs:677:60
    |
677 |                     box_once(paths.try_fold(v, |acc, path| path?.update(acc, &f)))
    |                                                            ^^^^^^^^^^^^^^^^^^^^^

error: `<D as DataT>::V<'_>` does not live long enough
   --> jaq-core/src/filter.rs:677:21
    |
677 |                     box_once(paths.try_fold(v, |acc, path| path?.update(acc, &f)))
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `jaq-core` due to 5 previous errors
~~~

Any ideas?

EDIT: I found that the errors above persist until (including) Rust 1.68 and disappear starting from Rust 1.69.

01mf02:
A little note: We currently have many type signatures in `jaq-std` looking like this:

~~~ rust
pub fn funs<D: DataT>() -> impl Iterator<Item = Filter<Native<D>>>
where
    for<'a> D::V<'a>: ValT,
{ ... }
~~~

This is a bit clunky. With [associated type bounds](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0/#bounds-in-associated-type-position), we could write this more compactly as follows, but that would increase MSRV to 1.79:

~~~ rust
pub trait DataTx: for<'a> DataT<V<'a>: ValT> {}
impl<T: DataT> DataTx for T where for<'a> T::V<'a>: ValT {}

pub fn funs<D: DataTx>() -> impl Iterator<Item = Filter<Native<D>>> { ... }
~~~

It's probably not a big deal for now, the whole added `where for<'a> ...` stuff amounts to less than 30 lines in jaq-std, which is quite bearable and probably not worth increasing the MSRV by at least 10 versions.

Issue #169, opened 2024-03-17
Parsing of zero-padded numbers


kklingenberg:
While parsing zero-padded numbers I came across this minor issue. This is a minimal example:

```
$ echo "0012" | cargo run .
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/jaq .`
0
0
12
```

Whereas jq yields just `12`.

This is `serde_json` at work, which in turn is probably following JSON's spec (is my guess). This is another view at the issue:

```
$ echo "0012" | cargo run -- -R fromjson 
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/jaq -R fromjson`
Error: cannot parse 0012 as JSON: end of file expected
```

Also, the lexer rejects these numbers too (which is fine, and consistent with the JSON parser). jq is also consistent with its lenient parser:

```
$ cargo run -- -n '0012'
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/jaq -n 0012`
Error: Unexpected token, expected as, *, +=, /=, %=, >=, /, ?, %, and, =, or, +, -, |, [, end of input, ==, -=, |=, //, *=, <=, ., !=, ,, >, <
   ╭─[<unknown>:1:2]
   │
 1 │ 0012
   │  ┬  
   │  ╰── Unexpected token 0
───╯

$ jq -n '0012'
12
```

Anyway, while attempting to work with these numbers one could hope to use the `tonumber` filter, but that's also implemented in terms of `fromjson`, so no luck there.

My suggestion is to either:

- document the non-leniency of the JSON parser, and the difference with jq's
- provide a `tonumber` filter that's more tolerant

kklingenberg:
An example of another side-effect of the current implementation of `tonumber`:

```
$ echo '"{}"' | jaq tonumber
{}
```

wader:
Related https://github.com/jqlang/jq/pull/3055 jq used to allow whitespaces for `tonumber` but not anymore

kklingenberg:
Interesting. So yet another side effect of `tonumber` just being `fromjson` is that it tolerates whitespace:

```
$ echo ' 12 ' | jaq -Rc '[., tonumber]'
[" 12 ",12]
```

pkoppstein:
@kklingenberg - Good catch re `jaq -n '"{}"|tonumber'`.  That's a bug that needs fixing.

[...]

PR #307, opened 2025-07-13
add toboolean/0 and error handling for conversions

djpohly:
Previously, passing a string representation of anything to `tonumber` would succeed; now it actually has to be a number or a string parseable as a number.

Added the missing `toboolean` while we're in the vicinity.

Includes tests.


01mf02:
Hi @djpohly, thanks for tackling this! This PR looks quite good, I just took the opportunity to make your filter definitions a bit more Haiku-style. :) Your code helped me a lot to realise the underlying pattern.

01mf02:
This also addresses #169.

PR #311, opened 2025-08-05
allow access to inner type of arena

slonkazoid:
this lets you allocate the code next to the AST so you can avoid lifetime hell

### DIFF
--- a/jaq-core/src/load/mod.rs
+++ b/jaq-core/src/load/mod.rs
@@ -11,6 +11,7 @@ use crate::{ops, path};
 #[cfg(feature = "std")]
 use alloc::boxed::Box;
 use alloc::{string::String, vec::Vec};
+use core::borrow::Borrow;
 pub use lex::Lexer;
 use lex::Token;
 pub use parse::Parser;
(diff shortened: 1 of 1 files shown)


01mf02:
Hi @slonkazoid! Before merging this, I would be interested in a concrete use case. Can you show me a code sample that profits from having this functionality?

slonkazoid:
i replied (twice) but email got blackholed i guess and github was 500ing 

i replied:

> cant exactly do that as github is currently 500ing for all PRs but i was parsing filters embedded in other strings, and keeping one arena which i can then pass into the filter parser made sense. issue was that the filter parser has to allocate a new `String` and it can't return the `Modules` type because the source String must live as long as it. giving it all the arena's lifetime made sense. however, i've moved onto doing the compilation in the filter parser so i no longer need this feature, but someone else might. looking at `typed_arena`'s api, having a shared reference to `Arena` doesn't let you deallocate or modify existing entries, so, it should not cause any problems

01mf02:
@slonkazoid, I just merged #372, so you can now access jaq's arenas. :)

PR #322, opened 2025-09-02
Use a tree-like structure instead of a linked list for variables.

01mf02:
This improves the complexity of variable lookup    from O(n) to O(log n), but
deteriorates  the complexity of variable insertion from O(1) to O(log n).

If the current O(n) complexity for variable lookup should turn out to be a bottleneck some day, this PR is the way to go.

### DIFF
--- a/jaq-core/src/filter.rs
+++ b/jaq-core/src/filter.rs
@@ -5,7 +5,7 @@ use crate::compile::{Bind, Fold, Pattern, Tailrec, Term as Ast, TermId as Id};
 use crate::data::{DataT, HasLut};
 use crate::fold::fold;
 use crate::val::{ValR, ValT, ValX, ValXs};
-use crate::{exn, rc_lazy_list, Bind as Arg, Error, Exn, RcList};
+use crate::{exn, rc_lazy_list, Bind as Arg, Error, Exn, RcList, RcTree};
 use alloc::boxed::Box;
 use dyn_clone::DynClone;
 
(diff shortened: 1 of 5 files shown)


01mf02:
This is probably not worth merging; the benchmark results show that this change deteriorates performance in many cases significantly:

|Benchmark|n|Before|After
|--|--|-:|-:
|`empty`|512|**330**|370
|`bf-fib`|13|440|**420**
|`defs`|100000|**50**|**50**
|`upto`|8192|**0**|**0**
|`reduce-update`|16384|**0**|10
|`reverse`|1048576|**40**|**40**
|`sort`|1048576|100|**90**
|`group-by`|1048576|**270**|**270**
|`min-max`|1048576|**160**|180
|`add`|1048576|**410**|450
|`kv`|131072|**90**|100
|`kv-update`|131072|**110**|120
|`kv-entries`|131072|**500**|520
|`ex-implode`|1048576|**470**|530
|`reduce`|1048576|**680**|740
|`try-catch`|1048576|170|**160**
|`repeat`|1048576|**150**|160
|`from`|1048576|**290**|**290**
|`last`|1048576|**20**|**20**
|`pyramid`|524288|**290**|300
|`tree-contains`|23|**70**|**70**
|`tree-flatten`|17|**750**|770
|`tree-update`|17|**430**|440
|`tree-paths`|17|**190**|**190**
|`to-fromjson`|65536|**40**|**40**
|`ack`|7|**420**|500
|`range-prop`|128|**350**|**350**
|`cumsum`|1048576|**240**|270
|`cumsum-xy`|1048576|**370**|390

PR #325, opened 2025-09-15
Add `sync` feature to make json::Val Sync+Send

I-Al-Istannen:
This allows moving it off the main thread, processing json in parallel and moving the results back to main.

I settled on `RefPtr` as it is quite short and a common name in other languages for reference counted pointers.

### DIFF
--- a/jaq-json/src/lib.rs
+++ b/jaq-json/src/lib.rs
@@ -47,10 +47,10 @@ pub use funs::base_funs;
 pub use funs::funs;
 pub use num::Num;
 
-#[cfg(feature = "sync")]
-pub(crate) type RefPtr<T> = alloc::sync::Arc<T>;
 #[cfg(not(feature = "sync"))]
-pub(crate) type RefPtr<T> = alloc::rc::Rc<T>;
+pub use alloc::rc::Rc;
+#[cfg(feature = "sync")]
+pub use alloc::sync::Arc as Rc;
 
 /// JSON superset with binary data and non-string object keys.
 ///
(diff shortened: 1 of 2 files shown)


jaq-json/src/lib.rs
```diff
@@ -512,12 +517,12 @@ impl core::ops::Add for Val {
                 Ok(Str(buf.into(), tag))
             }
             (Arr(mut l), Arr(r)) => {
-                //std::dbg!(Rc::strong_count(&l));
-                Rc::make_mut(&mut l).extend(r.iter().cloned());
+                //std::dbg!(Ptr::strong_count(&l));
```
01mf02, line 520: Shouldn't this be `RefPtr` here as well?

I-Al-Istannen: Yea, good catch. Probably died in some search&replace and did not show up as it is uncommented.

jaq-json/src/lib.rs
```diff
@@ -47,6 +47,11 @@ pub use funs::base_funs;
 pub use funs::funs;
 pub use num::Num;
 
+#[cfg(feature = "sync")]
```
01mf02, line 50: Can't we actually do `pub(crate) use alloc::sync::Arc as Rc` here and use `Rc` everywhere? That would make this PR much smaller ...

I-Al-Istannen: If you are fine with it, I certainly can. I just did not want to lie about what `Rc` refers to, as many people have expectations when reading it :) If you are want it though, I will gladly change it. I have no strong opinion.

01mf02:
Thanks again for your contribution! :)

I-Al-Istannen:
Thank you for ~~(implementing it)~~ merging and maintaining jaq :)

PR #337, opened 2025-09-29
Make `byteoffset` take two arguments.

01mf02:
This makes the usage of `byteoffset` (hopefully) more ergonomic and easier to remember.
Now, `byteoffset($x; $y)` calculates $o_x - o_y$, where $o_x$ and $o_y$ are the (byte) addresses of `$x` and `$y` in memory, respectively.
That means that `byteoffset/2` now looks more like a subtraction operation, which it actually also is.

### DIFF
--- a/examples/tar.jq
+++ b/examples/tar.jq
@@ -15,7 +15,7 @@
 # this decoder stores *where* that data came from.
 # That means that we can obtain e.g. the byte offset of each file header:
 #
-# . as $i | decode_tar | {name: .o.name, offset: .i | byteoffset($i)}
+#     . as $i | decode_tar | {name: .o.name, offset: byteoffset(.i; $i)}
 #
 # All these commands take about 200 milliseconds on my machine.
 # For comparison, `tar --list --file 1G.tar` takes about 10 milliseconds.
@@ -86,8 +86,7 @@ def decode_entry:
   . as $input |
   # we use i as remaining input, which we later use to infer consumed input
   {i: .} |
-  # TODO: make byteoffset fail
-  def consumed: .i | byteoffset($input) // error;
+  def consumed: byteoffset(.i; $input);
   def offset: (BLOCK_BYTES - (consumed % BLOCK_BYTES)) % BLOCK_BYTES;
   str("name"; 100) |
   oct("mode"; 8) |
@@ -104,7 +103,7 @@ def decode_entry:
 ;
 
 # Translate from i denoting remaining input to i denoting consumed input.
-def set_consumed($input): .i |= .[:byteoffset($input)];
+def set_consumed($input): .i |= .[:byteoffset(.; $input)];
 
 def decode_tar: tobytes |
   ([limit(BLOCK_BYTES*2; repeat(0))] | tobytes) as $END_MARKER |
--- a/jaq-json/tests/funs.rs
+++ b/jaq-json/tests/funs.rs
@@ -10,9 +10,9 @@ yields!(bsearch_absent2, "[1, 3] | bsearch(2)", -2);
 yields!(bsearch_absent3, "[1, 3] | bsearch(4)", -3);
 yields!(bsearch_present, "[1, 3] | [bsearch(1, 3)]", [0, 1]);
 
-yields!(byteoff1, r#""asd" as $x | $x[1:] | byteoffset($x)  "#, 1);
-yields!(byteoff2, r#""asd" as $x | "df" | byteoffset($x)? // 9"#, 9);
-yields!(byteoff3, r#""asd" as $x | $x | byteoffset($x[1:])"#, -1);
+yields!(byteoff1, r#""asd" | byteoffset(.[1:]; .)"#, 1);
+yields!(byteoff2, r#""asd" | byteoffset(.; .[1:])"#, -1);
+yields!(byteoff3, r#""asd" | byteoffset("df"; .)? // 9"#, 9);
 
 // 41 = 0x29
 yields!(fromcbor1, "[41] | tobytes | fromcbor", -10);
--- a/jaq-std/src/lib.rs
+++ b/jaq-std/src/lib.rs
@@ -495,8 +495,8 @@ where
         ("ascii_upcase", v(0), |cv| {
             bome(cv.1.map_utf8_str(ByteSlice::to_ascii_uppercase))
         }),
-        ("byteoffset", v(1), |mut cv| {
-            bome(byte_offset(cv.0.pop_var(), cv.1))
+        ("byteoffset", v(2), |mut cv| {
+            bome(byte_offset(cv.0.pop_var(), cv.0.pop_var()))
         }),
         ("reverse", v(0), |cv| bome(cv.1.mutate_arr(|a| a.reverse()))),
         ("keys_unsorted", v(0), |cv| {


pkoppstein:
If I understand things correctly, byteoffset/1 was written
in accordance with the general jq approach to filters and pipelines,
and in particular was closely related to index/1.
I therefore believe it should be retained, whether or not
byteoffset/2 is introduced.

wader:
Are there other builtins in jq that come in /1 and /2 variant? all math functions that take two args are only /2 i think?

01mf02:
> If I understand things correctly, byteoffset/1 was written in accordance with the general jq approach to filters and pipelines, and in particular was closely related to index/1. I therefore believe it should be retained, whether or not byteoffset/2 is introduced.

There are indeed many filters in jq that work like `index/1`, e.g. `startswith/1`, `trimstr/1`, `join/1`, `scan/1` etc.
However, I think that they are all used in a quite different way than `byteoffset`, namely frequently with a *constant* argument. For example, it is common to write `scan("[0-9]*")`, `trimstr("\n")`, `join(" ")`. However, you can *not* do this with `byteoffset` --- calling `byteoffset` with a constant as parameter will always yield an error. That's why in practice, the way you call `byteoffset` is different from these other functions.
At least that is my experience from using `byteoffset`.

I feel that there is also another difference, even if it is a bit more subtle: The aforementioned filters like `index/1` etc. more or less follow a similar convention, namely taking as input a "large" string (often coming from user data) and taking as argument a "small" string (often hard-coded). On the other hand, both arguments to `byteoffset` come from the same source (normally user data), and `byteoffset` treats both arguments mostly the same --- indeed, `byteoffset(x; y)` is the same as `-byteoffset(y; x)`. Such a relationship does not hold for any of the aforementioned filters. That's why `byteoffset` is special.

When I tried to use `byteoffset/1`, I was having quite a hard time to remember what had to go as input and what as argument. Furthermore, I was frequently forced to bind stuff to variables, which then got used only a single time. With `byteoffset/2`, both these problems went away.

01mf02:
> Are there other builtins in jq that come in /1 and /2 variant? all math functions that take two args are only /2 i think?

I am not aware of any such builtins where `$x | f($y)` is the same as `null | f($x; $y)`. There is only one variant for math functions, I believe.

pkoppstein:
@01mf02 - Thank you for the explanation.  I see now that I had misunderstood the writeup at https://github.com/01mf02/jaq/pull/284, because I had assumed that you were still adhering to (a) basic principles of referential transparency, and (b) the jq approach to `values` (according to which, every jq variable holds a value that is JSON or NaN or Inf).

On further reflection, though, I am wondering whether it wouldn't be possible (and indeed better) to "have your cake (i.e., taking advantage of Rust slices) and eat it (i.e., preserving referential transparency)".   Specifically, byteoffset/2 would have enough information to determine whether the O(1) computation can be performed; if not, it can fall back on the more usual jq semantics. 

(Notice also that if jaq were to approach `byteoffset` in this jq-conformant manner, one might as well provide it in as byteoffset/1 in a manner analogous to `index/1`.)

01mf02:
> On further reflection, though, I am wondering whether it wouldn't be possible (and indeed better) to "have your cake (i.e., taking advantage of Rust slices) and eat it (i.e., preserving referential transparency)". Specifically, byteoffset/2 would have enough information to determine whether the O(1) computation can be performed; if not, it can fall back on the more usual jq semantics.

What do you mean by the "more usual jq semantics"? What should `byteoffset/2` fall back to?

If you have in mind something like `index`, then I think that having such a "fall back" mechanism could lead to potentially very confusing situations.

Suppose that we would fall back to an `index/1`-like behaviour when the arguments of `byteoffset` do not overlap in memory. So something like `def byteoffset_or_index($x; $y): try byteoffset($x; $y) catch ($y | index($x))`.

Then consider the example: `"abcabc" | byteoffset_or_index(.[3:]; .)`. This will yield `3`, in constant time. However, `"abcabc" | byteoffset_or_index("abc"; .)` would return ... 0! Even if `.[3:]` in the first example is `"abc"`. That means that by replacing equal things, we suddenly get different outputs (whereas currently, we would get an error). Not only that, but we also get vastly different performance.

For me, `byteoffset` should do one thing, and it should do it predictably, namely: if two strings overlap in memory, then return their offsets in bytes, else fail with an error. That's it. By its very nature, this operation is not referentially transparent, but I think that if one is frank about it, that's no problem. On the other hand, I'm always critical about implicit fallback mechanisms, because they make the overall behaviour harder to understand, especially if things go wrong: In particular, I'm worried about the scenario where a user of `byteoffset` messes up things and tries to find byte offsets of non-overlapping strings, which however share common parts. In that case, a fallback mechanism might return some result, instead of failing right away, making it harder for the user to recognise that there should have been an error.

Finally, there is yet another case that makes a fallback mechanism problematic: Consider the example `"abcdef" | byteoffset(.[1:]; .[:-1])`. Here, `.[1:] == "bcdef"` and `.[:-1] == "abcde"`, so neither string is contained in the other, yet `byteoffset(...)` returns `1` in this case. Now you could imagine some fallback algorithm that also returns `1` here, but I have a bit of a hard time to come up with a comprehensible description of what this should actually do. And if I have such problems to explain such a thing, then it is a sign for me that the idea behind it is questionable.

01mf02:
By the way, I'm not sure whether we mean the same thing by "referential transparency". Would you like to elaborate on why you think that my proposed version of `byteoffset/2` breaks referential transparency?

01mf02:
However, I just thought about another thing that is slightly disturbing me; `byteoffset` breaks round-trippability.
That is, I thought that unlike jq, `jaq '(f) | g'` is equivalent to `jaq 'f' | jaq 'g'` (assuming that both `f` and `g` are free of side-effects like `inputs`). However, with `byteoffset`, that property does not hold anymore, because the information about which slice is shared with another one is lost by serialisation. That is a bit of a bummer, considering that I have introduced the extended JSON format notably to have more robust round-tripping guarantees.

pkoppstein:
@01mf02 wrote:

> By the way, I'm not sure whether we mean the same thing by "referential transparency".

In the context of jq, I believe referential transparency can be
said to boil down to the principle that, using usual jq notation:

  if $a == $b and if E is a jq expression involving $a (and possibly $b), and
  if F is the same expression but with all occurrences of $a replaced by $b,
  then E == F.  

Note that E could be a jq pipeline (i.e. a jq expression involving `|`).

01mf02:
> if $a == $b and if E is a jq expression involving $a (and possibly $b), and
> if F is the same expression but with all occurrences of $a replaced by $b,
> then E == F.

In that case, jq does not uphold referential transparency, because:

~~~
$ jq -n '{a: 1, b: 2} as $a | {b: 2, a: 1} as $b | $a == $b, ($a | keys_unsorted) == ($b | keys_unsorted)'
true
false
~~~

However, I believe that jq does guarantee something different:
For two filters `f` and `g`, if both `f` and `g` neither yield an error nor have side effects, and `f` does not yield `NaN` or `Infinity`, then `jq '(f) | g'` is equivalent to `jq 'f' | jq 'g'`. And *this* property would be broken by jaq, if `byteoffset` was available. That's why I have come to think that `byteoffset` is a bad idea.

I have made a PR (#340) that removes `byteoffset` and makes the example TAR decoder work without `byteoffset`. Surprisingly, the resulting code is not that much more involved than with `byteoffset`, even if it is a bit slower.

pkoppstein:
@01mf02 - Yes, my brief synopsis of what referential transparency would look like in jq was over-simplified: there must be some restrictions on E (`now` is another complication).  Anyway, I understand that you have decided to remove the problematic version of "byteoffset" altogether.  Thanks again for the reconsideration.

01mf02:
> @01mf02 - Yes, my brief synopsis of what referential transparency would look like in jq was over-simplified: there must be some restrictions on E (`now` is another complication). Anyway, I understand that you have decided to remove the problematic version of "byteoffset" altogether. Thanks again for the reconsideration.

Oh dear, you're right; I didn't even think about `now`. Good to keep that in mind.

I have to thank you again for your very precious feedback. In hindsight, `byteoffset` was not such a great idea. :P So thanks again for having stepped in. :)

Issue #323, opened 2025-09-11
Parallel processing with `jaq_json`


I-Al-Istannen:
Hey,

I am currently using the Rust API for trying to process slightly larger datasets (a few thousand json files with a few hundred MB total) with a few queries (100+). As this problem is basically embarrassingly parallel, I parallelized my program over the queries, with <cpu count> worker threads each polling queries from a queue, running the query over all inputs and then collecting the results. This leads to the following flow:
1. Read all input files and queries
2. Parse all input files to json (`jaq_json`'s `Val`)
3. Clone this into each worker thread, so they have the inputs lying around
4. In each worker:
    1. Pick a query to run from the queue
    2. Clone your inputs, run the query against it, keep the result
    3. Loop
5. Collect the results from all queries and return it

This breaks completely at step 3 and 5, as `Val` is not `Send` due to the `Rc` and can not be passed to a different thread. Even if I move the json strings over, and re-parse everything in each worker thread (which takes quite a while…), this problem persists: I can not get the results back from the worker thread!

My first solution was defining my own JSON type that I could convert from and to `jaq_json`'s `Val` type, but the conversion and dropping costs of that approach lead to an order of magnitude slower execution.
My current solution basically copies `jaq_json` and replaces every `Rc` with `Arc`.

Could I talk you into defining a
```rs
#[cfg(feature = "sync-val")]
type Rc = Arc
#[cfg(not(feature = "sync-val"))]
type Rc = Rc
```
or something similiar to allow users to define this? As cargo features are additive, users might pay a small penalty if any crate in their graph uses parallel processing, but that sounds acceptable to me.

I can PR a feature like that if you want.

01mf02:
Hi @I-Al-Istannen, thanks for explaining your workflow so well! I am sympathetic to changes that help with using `Val` in a multi-threaded setting.

Your `#[cfg(...)]` approach seems like a viable route towards this. I was thinking alternatively to make `Val` abstract over the type of pointer à la [`PointerFamily`](https://rust-lang.github.io/rfcs/1598-generic_associated_types.html#associated-type-constructors-of-type-arguments), but I think that this might become a parameter fest quite fast.

Do you know of any other projects that have a similar build-time feature, in order to get inspiration for the feature name?

I-Al-Istannen:
[...]

PR #332, opened 2025-09-25
Manual & man page

01mf02:
This PR adds a manual for jaq.
It features tests that are extracted and run automatically by `make tests` (or plain `make`).

This PR adds a few new filters:

- `setpath/2`
- `delpaths/1`
- `pick/1`
- `halt/1` (not in `jq`)

Furthermore, it corrects CBOR aliases/anchors behaviour and makes jaq fail when trying to load CBOR undefined values.

PR #338, opened 2025-09-29
doc: Add a few cross references

I-Al-Istannen:
See https://github.com/01mf02/jaq/issues/323#issuecomment-3333517132.

I am not really happy with the link to `filter` on GitHub, but I don't think I can reference the binary crate in any useful way.


01mf02:
Thanks a lot, @I-Al-Istannen! Sorry it took me so long, I was pretty busy with documentation myself (#332). ;)

I agree with you about the link to `jaq::filter` on GitHub, that's why I removed it. Unfortunately, I did not get inter-crate documentation links to work.

PR #342, opened 2025-10-14
Use `regex` instead of `regex-lite`

01mf02:
This is an experiment, see [#1274](https://github.com/rust-lang/regex/pull/1274).

I evaluated jqjq with it as follows:

~~~
cd jaq
cargo build --release
git clone https://github.com/wader/jqjq
cd jqjq
git checkout 535076f3c4ede19817d5946a6d1703b04d18778a
time ../target/release/jaq -n -L . 'include "jqjq"; eval(".")'
~~~

This yields 0.921s with `regex` + PikeVM vs. 0.493s with `regex_lite`.
(For comparison, jq takes 0.499s.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -428,6 +428,7 @@ dependencies = [
  "jaq-json",
  "libm",
  "log",
+ "regex-automata",
  "regex-lite",
  "serde_json",
  "urlencoding",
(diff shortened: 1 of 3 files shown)


01mf02:
This is too slow, and because `regex-lite` better fits jaq's philosophy, let's keep it.

PR #348, opened 2025-10-20
Fix compilation with `cbor` feature without `formats`

jakobhellermann:
I was testing a PR I'm gonna open with `cargo hack check --each-feature` (https://github.com/taiki-e/cargo-hack) and noticed that compiling with `--feature cbor` but not `formats` fails due to `invalid_data` and `BoxError` missing.

This fixes that by unconditionally compiling them, since they're so small anyways.
Alternatively you could do `#[cfg(any(feature = "cbor", feature = "yaml", ...)]` or whatever the exact set of formats is that needs these items.

### DIFF
--- a/jaq-json/src/lib.rs
+++ b/jaq-json/src/lib.rs
@@ -728,10 +728,10 @@ impl fmt::Display for Val {
     }
 }
 
-/// Type alias for a type-erased Error
+/// Dynamic & thread-safe [`std::error::Error`].
 pub type BoxError = Box<dyn std::error::Error + Send + Sync>;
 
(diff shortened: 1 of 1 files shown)


01mf02:
I'm a bit critical towards `dead_code`, because then I might forget removing `invalid_data` if at some point I replace that with something else. However, I would be fine with:

- `#[cfg(any(feature = "cbor", feature = "json")]` for `invalid_data`
- `#[cfg(feature = "cbor")]` for `BoxError`

Or, even better: Make `invalid_data` and `BoxError` public, and use that inside `jaq/src/main.rs`. Yes, I think that would be a better solution.

01mf02:
Thanks again!

PR #350, opened 2025-10-20
Make `jaq_json::Map` public

jakobhellermann:
~~This makes some of the `as_x` accessors on `Value` and `Num` public, I've found them to be useful working on https://github.com/jakobhellermann/uniscan and thing it would be a good thing to have available as a user of the library.~~

EDIT: Just make `jaq_json::Map` public. It's useful when you want to construct a new object.

### DIFF
--- a/jaq-json/src/lib.rs
+++ b/jaq-json/src/lib.rs
@@ -134,7 +134,7 @@ impl Type {
 }
 
 /// Order-preserving map
-type Map<K = Val, V = K> = indexmap::IndexMap<K, V, foldhash::fast::RandomState>;
+pub type Map<K = Val, V = K> = indexmap::IndexMap<K, V, foldhash::fast::RandomState>;
 
 /// Error that can occur during filter execution.
 pub type Error = jaq_core::Error<Val>;


jaq-json/src/lib.rs
```diff
@@ -134,7 +134,7 @@ impl Type {
 }
 
 /// Order-preserving map
-type Map<K = Val, V = K> = indexmap::IndexMap<K, V, foldhash::fast::RandomState>;
+pub type Map<K = Val, V = K> = indexmap::IndexMap<K, V, foldhash::fast::RandomState>;
```
jakobhellermann: This makes it easier to construct objects from scratch:
```rs
let mut obj = jaq_json::Map::default();
obj.insert("file".to_string().into(), pptr_file.into());
obj.insert("path_id".to_string().into(), path_id.into());
obj.insert(
    "class_id".to_string().into(),
    format!("{class_id:?}").into(),
);
jaq_json::Val::Obj(Rc::new(obj))
```

jakobhellermann:
I just realized some of these methods are already available on `ValT`. Maybe this isn't necessary then

jakobhellermann:
Yup, that works https://github.com/jakobhellermann/uniscan/commit/bc7dd40f1eee9e1e58aec5d65753c51a94637b88

I'd still like `Map` to public, for constructing new objects from scratch:
```rs
let mut obj = jaq_json::Map::default();
obj.insert("file".to_string().into(), pptr_file.into());
obj.insert("path_id".to_string().into(), path_id.into());
obj.insert(
    "class_id".to_string().into(),
    format!("{class_id:?}").into(),
);
jaq_json::Val::Obj(Rc::new(obj))
```

Any maybe an `PartialEq` impl of `Val` to `&str` would be nice? On the other hand that opens a can of worms, because you probably don't want `PartialEq` implementations for every numeric type as well.

01mf02:
Having `Map` public is a good idea. Thanks for your initiative.

I'm more critical towards `PartialEq` for `Val` and `&str`, in particular because then why not have just all combinations of `PartialEq`, which would amount to `n²` `PartialEq` instances? I think it's better to keep it at a single instance.

PR #360, opened 2025-11-05
Prepare 3.0-alpha

01mf02:


pkoppstein:
@01mf02 - The best I’ve been able to come up with is what I wrote before:

if f and g are two jq filters, then the
filter `(f), (g)` emits a stream consisting of the items produced by
f followed by the items produced by g. If f is an atomic
expression, then (f) can be replaced by f, and likewise for g.

01mf02:
@pkoppstein, I appreciate your explanation, but I have to confess that it confuses me quite a bit, and it is quite far away both from what is implemented in jaq as well as my intuition.

I propose that we take a step back for a moment. Execution in jaq takes several steps:

~~~
String --lex--> [Token] --parse--> AST --run--> Values
~~~

So from a filter such as `1, 2 | 3`, the lexer creates tokens `["1", ",", "2", "|", "3"]`. From that, the parser creates an AST such as `Pipe(Comma(1, 2), 3)`. To create the AST, the parser considers the precedences of the operators. Finally, if we skip the compilation step for performance and name resolution, we could execute the AST directly, yielding a stream of values.

The important point here is that at the point where the manual defines *semantics*, i.e. for `f, g`, the filters `f` and `g` refer to *ASTs*, not to tokens. That means that at this point, there can be no more confusion as to whether `1, 2 | 3` is equivalent to `1, (2 | 3)` or `(1, 2) | 3`, because precedences have already been considered in the step before. The AST does not even have any notion of parentheses.

My explanations for semantics in the manual, such as for `f, g`, do not hold anymore if you consider them on the level of tokens (like you did), or, to give an extreme example, on the level of (string) characters. For example, you could equally well say that I did not define the semantics of `f  ,   g`, because I did not specify that whitespace is insignificant. This is a fair point --- the manual leaves a few things implicit, for the sake of being not too verbose.

The interesting question is: Should this be clarified? And if so, how to do it so that it does not confuse users more than it clarifies affairs?

01mf02:
I would also appreciate it if you could open an issue to continue this conversation. Other people might also have some good ideas to contribute.

PR #361, opened 2025-11-05
chore(docs): fix manpage generation

tranzystorekk:
pandoc's manpage format expects the manual section to be passed as a separate variable: https://pandoc.org/MANUAL.html#variables-for-man-pages


01mf02:
Thanks @tranzystorekk, great catch!

PR #368, opened 2025-11-10
Slice strings/arrays when indexing with objects.

### LINKED ISSUE #367: Possibly support array index using range object?
I noticed this when playing around with jqjq and jaq

```sh
$ jq -cn '[1,2] | .[{start:0, end:1}]'
[1]

$ jaq --version
jaq 3.0.0-alpha
$ jaq -cn '[1,2] | getpath([{start:0, end:1}])'
[1]
$ jaq -cn '[1,2] | .[{start:0, end:1}]'
Error: cannot index [1,2] with {"start":0,"end":1}
```

01mf02:
This PR enables slicing via indexing with `{start, end}` objects. For example:

- `[1, 2, 3] | .[{start: 1}] --> [2, 3]`
- `[1, 2, 3] | .[{end: -1}] --> [1, 2]`
- `[1, 2, 3] | .[{start: 1, end: -1}] --> [2]`

Furthermore, this also allows using `null` as argument to the slicing operator:

- `[1, 2, 3] | .[1:null] --> [2, 3]`
- `[1, 2, 3] | .[null:-1] --> [1, 2]`
- `[1, 2, 3] | .[null:null] --> [1, 2, 3]`

This is currently a bit more permissive than `jq`, because indexing an array with _any_ object yields the original array; i.e. `[1, 2, 3] | .[{}] --> [1, 2, 3]`. That is because an empty array is interpreted as `{start: null, end: null}`. Here, jq throws an error.

This should close #367.

### DIFF
--- a/jaq-json/src/lib.rs
+++ b/jaq-json/src/lib.rs
@@ -185,6 +185,9 @@ impl jaq_core::ValT for Val {
                 .as_pos_usize()
                 .and_then(|i| abs_index(i, a.len()))
                 .map_or(Val::Null, |i| a[i].clone())),
+            (v @ (Val::Str(..) | Val::Arr(_)), Val::Obj(o)) => {
+                v.range(o.get(&Val::utf8_str("start"))..o.get(&Val::utf8_str("end")))
+            }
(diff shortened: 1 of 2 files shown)


wader:
https://github.com/wader/jqjq/pull/35 works with this change 🥳

Issue #374, opened 2025-11-21
Provide a more complete example for jaq_core


the80srobot:
The [example code](file:///Users/adam/.pizza/repo/scripts/helpers/target/doc/jaq_core/index.html) in jaq_core is incomplete and not practically usable:

* It doesn't load the standard library (and basically tells you to figure it out yourself)
* It doesn't explain concepts like Arena, or what the loader does, etc
* It doesn't say where errors can arise at runtime and how to catch them

Just a tiny bit more explanation would go a long way.

01mf02:
I agree with you. Would you like to contribute by expanding the example code / documentation in a way that you see fit?

the80srobot:
Hi! I actually gave up on that approach and what I did instead is basically clone the entire `jaq` crate, removing the `main` function and calling `real_main` directly as needed.

It's a little hard to tell with things changing now around 3.0.0-alpha, but it looks like you might need a larger refactor to make the language embeddable the way e.g. lua is.

But it looks to me like that refactor might actually just be:

* Make a crate called something like `jaq-embed`
* Do what I did, remove `main` and turn `real_main` into the entry point.
* Rename the Cli struct into something like Options and make it slightly nicer to use programmatically
* Then just change `jaq` to be a layer on top that

Does that seem completely off and like I missed something about the architecture? Or does it sound reasonable?

the80srobot:
Poking around the internals a little more, I think two current design decisions are really limiting.

- One is the assumption that the filter will output directly to stdout. This is mostly just baked into `jaq` and not any of the other crates, and could probably be changed so that the API takes a impl Write.
- The second one is harder to change and that is the decision to basically not support --slurp. (It's kind of supported, but doesn't combine the inputs into one array, so it's basically the same as calling the filter multiple times.) This might make sense for the CLI, but it's complicating when you're trying to use it programmatically.

So to amend, I think the intermediate crate `jaq-embed` would need the following modifications:

- Take all the options normally provided on the `Cli` + an `impl Write` for output. The `Cli` struct is already fairly easy to use, it would need only a few minor conveniences, like maybe automatically serializing struct arguments from Rust.
- Allow real slurping of inputs, so that batched processing is possible.

I think then you'd end up with an API that can run a `jq` program programmatically with real ease and allow you to capture output without blocking off stdin globally.

the80srobot:
(In case it's not obvious, I am offering to do that when I have a spare weekend.)

01mf02:
[...]

Issue #99, opened 2023-07-27
Example showing how to use jaq as a library is incomplete


kimjarvis:
I would like to use the crate as a library.  The example in the code comments of jaq-core/lib.rs does not compile.

```
15 |     let (f, errs) = jaq_parse::parse(filter, jaq_parse::main());
   |                     ^^^^^^^^^ use of undeclared crate or module `jaq_parse`
```

Also, the jaq crate combines the component crates into a workspace.  It is not clear what I should put in my Cargo.toml file to use this crate as a library.

01mf02:
Happy to hear that you want to use jaq as a library.

The code that you are referring to is from the current development version on Git, which has a quite different API than the latest stable release (0.10.1).
Code for using the stable release can be found at: https://docs.rs/jaq-core/latest/jaq_core/

You can browse the source code for the stable release at: https://github.com/01mf02/jaq/tree/v0.10

And finally, your Cargo.toml should include the following entries to use jaq:

~~~
jaq-core = { version = "0.10.1" }
jaq-std  = { version = "0.10.0" }
~~~

I hope that this works for you.

kimjarvis:
Thank you for your guidance.  I was able to get the example code in jaq-core/lib.rs to run.  I checked out the jaq master branch.  I added the path to these crates to my Cargo.toml file. 

Cargo.toml
```toml
[dependencies]
serde_json = "1.0"
jaq-core = { path = "../jaq/jaq-core" }
jaq-parse = { path = "../jaq/jaq-parse" }
```

src/main.rs
```rust
use jaq_core::{Ctx, FilterT, ParseCtx, RcIter, Val};
use serde_json::{json};

fn main() {

    let input = json!(["Hello", "world"]);
    let filter = ".[]";

    // start out only from core filters,
    // which do not include filters in the standard library
    // such as `map`, `select` etc.
    let mut defs = ParseCtx::new(Vec::new());

    // parse the filter
    let (f, errs) = jaq_parse::parse(filter, jaq_parse::main());
    assert_eq!(errs, Vec::new());

    // compile the filter in the context of the given definitions
    let f = defs.compile(f.unwrap());
    assert!(defs.errs.is_empty());

    let inputs = RcIter::new(core::iter::empty());

    // iterator over the output values
    let mut out = f.run((Ctx::new([], &inputs), Val::from(input)));

    assert_eq!(out.next(), Some(Ok(Val::from(json!("Hello")))));
[...]

PR #375, opened 2025-11-26
Open the API & improve pretty-printing

01mf02:
This PR exposes several previously hidden (and potentially duplicated) APIs as part of the `jaq-all` crate, such as:

- compiler diagnostics
- pretty printing
- value reading/writing

This should make it much easier to embed jaq into your applications (as requested in #374, but also previously in #99), while still profiting from nice error messages, multi-format support etc.

Furthermore, it adds support for `JQ_COLORS` to the `jaq` CLI.

As a small change, `halt_error` now does not yield its input anymore.

# YAML pretty-printing

This PR also vastly improves pretty printing for YAML --- for example, the output of `jaq -n '[1, {a: 2, b: [3]}]' --to yaml` is now:

~~~
---
- 1
-
  a: 2
  b:
    - 3
...
~~~

To omit the `---`/`...` markers for YAML document start/end, you can now use `-j` (`--join-output`).
The compact output for YAML (via `-c` or `toyaml`) has also been adapted to use spaces after commas and colons:

~~~
$ jaq -n '[1, {a: 2, b: [3]}]' --to yaml -cj
[1, {a: 2, b: [3]}]
~~~

This change was necessary because object keys such as `"a"` are now printed as `a`, therefore printing `a:1` would be interpreted as `"a:1"`. For that reason, it becomes necessary to print spaces after colons, and for consistency, it is less confusing to have spaces after commas too, even if they are not strictly necessary per se.

### DIFF
--- a/jaq-all/examples/main.rs
+++ b/jaq-all/examples/main.rs
@@ -1,4 +1,4 @@
-use jaq_all::{data, load, fmts};
+use jaq_all::{data, fmts, load};
 use std::io::{self, Error, ErrorKind};
 
 fn main() -> io::Result<()> {
--- a/jaq-json/src/lib.rs
+++ b/jaq-json/src/lib.rs
@@ -32,9 +32,7 @@ use jaq_core::{load, ops, path, val, Exn};
 use num_bigint::BigInt;
 use num_traits::{cast::ToPrimitive, Signed};
 
-// TODO!!!
-pub use funs::bytes_valrs;
-pub use funs::funs;
+pub use funs::{bytes_valrs, funs};
 pub use num::Num;
 
 #[cfg(not(feature = "sync"))]


wader:
"bla" as in bla bla bla? 😬🤔

01mf02:
> "bla" as in bla bla bla? 😬🤔

"bla" as in "Better Label Arriving". ;)

I'm not sure yet what exactly should go into this new crate. At the beginning, this was rather small, but now it turns out that extra format support finds a good home here. Just brainstorming: `jaq-full`, `jaq-all`, `jaq-extra`, ...

syyyr:
Hi, I am trying to use the code from the example in `jaq-all/examples/main.rs` to use jaq with my custom type. However, the `jaq_all::data::run` function only accepts Val. Should I use a different function? I have implemented ValT for my type.

01mf02:
> Hi, I am trying to use the code from the example in `jaq-all/examples/main.rs` to use jaq with my custom type. However, the `jaq_all::data::run` function only accepts Val. Should I use a different function? I have implemented ValT for my type.

Ahoj @syyyr! Yes, if you do not use `jaq_json::Val`, you will need to run your filters with [`filter.id.run`](https://docs.rs/jaq-core/latest/jaq_core/trait.FilterT.html#tymethod.run) directly, like `jaq_all::data::run` does it internally. You might still wish to use `jaq_all::compile_with`, though.

syyyr:
Okay, it seems that I have no idea what I'm doing: I have implemented ValT for my type, but it seems that to use filter.id.run, I need to supply some sort of Ctx. There are some structs like that for Val, but not for generic ValT types. Am I supposed to create my own context? Then there's the HasLut traits and others, should I do something with those? I tried just copying the code inside of jaq_all::data::run, but it needs all that.

syyyr:
Okay, I have somehow managed to get some results: https://github.com/silicon-heaven/libshvproto-rs/commit/bc74965e8f3dc544a30cc08a22b97a3005d2a2bf

Though I still think I have too much code. Anyway, thanks for the help, I won't spam this PR anymore. :)

01mf02:
> Okay, it seems that I have no idea what I'm doing: I have implemented ValT for my type, but it seems that to use filter.id.run, I need to supply some sort of Ctx. There are some structs like that for Val, but not for generic ValT types. Am I supposed to create my own context? Then there's the HasLut traits and others, should I do something with those? I tried just copying the code inside of jaq_all::data::run, but it needs all that.

This is explained in [`Ctx::new`](https://docs.rs/jaq-core/3.0.0-beta/jaq_core/struct.Ctx.html#method.new). In a nutshell, if you do not need the `inputs` filter, you should do fine with `JustLut` as your `Data` type. Otherwise, you need to define your own `Data` type.

> Okay, I have somehow managed to get some results: [silicon-heaven/libshvproto-rs@bc74965](https://github.com/silicon-heaven/libshvproto-rs/commit/bc74965e8f3dc544a30cc08a22b97a3005d2a2bf)
> 
> Though I still think I have too much code. Anyway, thanks for the help, I won't spam this PR anymore. :)

That looks good, @syyyr. Still, I think that you do not need the `runner`. The Rust compiler would also have told you that if you would have omitted the `pub` on most of your data types. ;)

Issue #422, opened 2026-03-28
Compound operator chaining does not support interspersed dots


CosmicToast:
Consider the following object:
`{"package": {"foo": "bar"}}`

Then consider the queries seeking to get out `bar`, in jq and jaq respectively.

In jq: `.package.foo` and `.package.["foo"]` both return `bar`.
In jaq, `.package.foo` returns `bar`, but `.package.["foo"]` complains about an invalid token.
`.package["foo"]` works as expected, but is non-obvious (relative to `.["package"]["foo"]`; note that `.["package"].["foo"]` is similarly valid in jq but not jaq).

wader:
For reference why it was added in jq this comment probably summaries it https://github.com/jqlang/jq/issues/1699#issuecomment-1624253094. For strict jq compat it would be good to support it, if not maybe make jaq produce clearer error message about it? currently: `Error: expected key <...inline code...> unexpected token`

CosmicToast:
From a UX perspective, `.foo.bar` and `.foo.["bar"]` being equivalent makes sense, since it's symmetrical.
You write in the `.foo.bar` format first, and then simply add `["` and `"]` if you realize you need those for compatibility reasons (in my case, `+@` at the start). Needing to take extra steps like removing the `.` as well is not symmetrical and therefore surprising, so if it's to stay that way I would think it should also warrant a compatibility note and a warning around the part of the manual that talks about `[]` notation and operator chaining.

01mf02:
Bonjour @CosmicToast,

thanks for reporting this. I think that #423 should address this.

Greetings to Alsace!

PR #423, opened 2026-03-30
Support compound paths with non-key heads

01mf02:
This addresses #422.

TODO:

- [x] Update documentation
- [x] Add examples


CosmicToast:
Tested on my non-trivial example (epub metadata extractor), looks good from my end :)

PR #432, opened 2026-04-29
Add `input_filename` support

### LINKED ISSUE #144: support `input_filename`
jq documents input_filename as follows: https://jqlang.github.io/jq/manual/#input_filename
> Returns the name of the file whose input is currently being filtered. Note that this will not work well unless jq is running in a UTF-8 locale.

jaq returns an error instead.
```
Error: undefined filter
   ╭─[<unknown>:1:35]
   │
 1 │ select(.name == "asdfasdfasdf") | input_filename
   │                                   ───────┬──────
   │                                          ╰──────── undefined filter
───╯
```

a-n-d-r-e-w-l:
Adds the `input_filename` filter from `jq` in the manner suggested [here](https://github.com/01mf02/jaq/issues/144#issuecomment-4343283722) (binding a variable and using normal filters to expose it).

Also adds a `input_filename/1` version that can override the value returned when filtering data on stdin; `jq` returns "<stdin>" in that case, which cannot be differentiated from filtering a _file_ called "<stdin>" - a rare occurrence, sure, but it costs almost nothing to add handling for that. The `/0` variant - which is the only variant that `jq` itself implements - is compatible with the `jq` version in all the circumstances I could think to test.

This implementation adds the variable fallback handling in an ad-hoc manner in `jaq-core`; it may make more sense to only define it in the main `jaq` crate (the CLI), but this approach provides the most flexibility both to CLI users and to people using the Rust crates directly.

Closes: #144 

### Checklist

- [X] I have read the [contribution guidelines](../CONTRIBUTING.md).

### DIFF
--- a/jaq/tests/golden.rs
+++ b/jaq/tests/golden.rs
@@ -1,14 +1,6 @@
 use std::{env, io, process, str};
 
-fn golden_test(args: &[&str], input: &str, out_ex: &str) -> io::Result<()> {
-    let mut child = process::Command::new(env!("CARGO_BIN_EXE_jaq"))
-        .args(args)
-        .stdin(process::Stdio::piped())
-        .stdout(process::Stdio::piped())
-        .spawn()?;
-
-    use io::Write;
-    child.stdin.take().unwrap().write_all(input.as_bytes())?;
+fn test_expected(child: process::Child, out_ex: &str) -> io::Result<()> {
     let output = child.wait_with_output()?;
     assert!(output.status.success());
 
@@ -18,9 +10,35 @@ fn golden_test(args: &[&str], input: &str, out_ex: &str) -> io::Result<()> {
     if out_ex.trim() != out_act.trim() {
         println!("Expected output:\n{}\n---", out_ex);
         println!("Actual output:\n{}\n---", out_act);
-        process::exit(2);
+        return Err(io::Error::new(
+            io::ErrorKind::Other,
+            "incorrect test output",
+        ));
     }
     Ok(())
+}
+
+fn golden_test(args: &[&str], input: &str, out_ex: &str) -> io::Result<()> {
+    let mut child = process::Command::new(env!("CARGO_BIN_EXE_jaq"))
+        .args(args)
+        .stdin(process::Stdio::piped())
+        .stdout(process::Stdio::piped())
+        .spawn()?;
+
+    use io::Write;
+    child.stdin.take().unwrap().write_all(input.as_bytes())?;
+
+    test_expected(child, out_ex)
+}
+
+fn test_no_input(args: &[&str], out_ex: &str) -> io::Result<()> {
+    let child = process::Command::new(env!("CARGO_BIN_EXE_jaq"))
+        .args(args)
+        .stdin(process::Stdio::null())
+        .stdout(process::Stdio::piped())
+        .spawn()?;
+
+    test_expected(child, out_ex)
 }
 
 macro_rules! test {
@@ -182,4 +200,48 @@ test!(
     &["-c", "-L", "tests", r#"include "a"; [a, data, d]"#],
     "0",
     r#"["bcddd",[1,2],3]"#
+);
+
+#[test]
+fn input_filename_file() -> io::Result<()> {
+    test_no_input(
+        &["--slurp", "true | input_filename", "./tests/data.json"],
+        r#""./tests/data.json""#,
+    )
+}
+
+#[test]
+fn input_filename_override() -> io::Result<()> {
+    test_no_input(
+        &[
+            "--slurp",
+            "--argjson",
+            "input_filename",
+            "0",
+            "true | input_filename",
+            "./tests/data.json",
+        ],
+        r#"0"#,
+    )
+}
+
+test!(
+    input_filename_stdin,
+    &["true | input_filename"],
+    "0",
+    r#""<stdin>""#
+);
+
+test!(
+    input_filename_stdin_variable,
+    &["true | $input_filename"],
+    "0",
+    r#""""#
+);
+
+test!(
+    input_filename_null_input,
+    &["-n", "true | input_filename"],
+    "",
+    r#"null"#
 );
--- a/jaq/tests/golden.rs
+++ b/jaq/tests/golden.rs
@@ -1,6 +1,15 @@
 use std::{env, io, process, str};
 
-fn test_expected(child: process::Child, out_ex: &str) -> io::Result<()> {
+fn golden_test(args: &[&str], input: &str, out_ex: &str) -> io::Result<()> {
+    let mut child = process::Command::new(env!("CARGO_BIN_EXE_jaq"))
+        .args(args)
+        .stdin(process::Stdio::piped())
+        .stdout(process::Stdio::piped())
+        .spawn()?;
+
+    use io::Write;
+    child.stdin.take().unwrap().write_all(input.as_bytes())?;
+
     let output = child.wait_with_output()?;
     assert!(output.status.success());
 
@@ -16,29 +25,6 @@ fn test_expected(child: process::Child, out_ex: &str) -> io::Result<()> {
         ));
     }
     Ok(())
-}
-
-fn golden_test(args: &[&str], input: &str, out_ex: &str) -> io::Result<()> {
-    let mut child = process::Command::new(env!("CARGO_BIN_EXE_jaq"))
-        .args(args)
-        .stdin(process::Stdio::piped())
-        .stdout(process::Stdio::piped())
-        .spawn()?;
-
-    use io::Write;
-    child.stdin.take().unwrap().write_all(input.as_bytes())?;
-
-    test_expected(child, out_ex)
-}
-
-fn test_no_input(args: &[&str], out_ex: &str) -> io::Result<()> {
-    let child = process::Command::new(env!("CARGO_BIN_EXE_jaq"))
-        .args(args)
-        .stdin(process::Stdio::null())
-        .stdout(process::Stdio::piped())
-        .spawn()?;
-
-    test_expected(child, out_ex)
 }
 
 macro_rules! test {
@@ -202,46 +188,23 @@ test!(
     r#"["bcddd",[1,2],3]"#
 );
 
-#[test]
-fn input_filename_file() -> io::Result<()> {
-    test_no_input(
-        &["--slurp", "true | input_filename", "./tests/data.json"],
-        r#""./tests/data.json""#,
-    )
-}
-
-#[test]
-fn input_filename_override() -> io::Result<()> {
-    test_no_input(
-        &[
-            "--slurp",
-            "--argjson",
-            "input_filename",
-            "0",
-            "true | input_filename",
-            "./tests/data.json",
-        ],
-        r#"0"#,
-    )
-}
+test!(
+    input_filename_file,
+    &["--slurp", "input_filename", "./tests/data.json"],
+    "",
+    r#""./tests/data.json""#
+);
 
 test!(
     input_filename_stdin,
-    &["true | input_filename"],
+    &["input_filename"],
     "0",
     r#""<stdin>""#
 );
 
 test!(
-    input_filename_stdin_variable,
-    &["true | $input_filename"],
-    "0",
-    r#""""#
-);
-
-test!(
     input_filename_null_input,
-    &["-n", "true | input_filename"],
+    &["-n", "input_filename"],
     "",
     r#"null"#
 );
--- a/jaq/tests/golden.rs
+++ b/jaq/tests/golden.rs
@@ -187,24 +187,3 @@ test!(
     "0",
     r#"["bcddd",[1,2],3]"#
 );
-
-test!(
-    input_filename_file,
-    &["--slurp", "input_filename", "./tests/data.json"],
-    "",
-    r#""./tests/data.json""#
-);
-
-test!(
-    input_filename_stdin,
-    &["input_filename"],
-    "0",
-    r#""<stdin>""#
-);
-
-test!(
-    input_filename_null_input,
-    &["-n", "input_filename"],
-    "",
-    r#"null"#
-);
--- a/jaq/tests/golden.rs
+++ b/jaq/tests/golden.rs
@@ -9,7 +9,6 @@ fn golden_test(args: &[&str], input: &str, out_ex: &str) -> io::Result<()> {
 
     use io::Write;
     child.stdin.take().unwrap().write_all(input.as_bytes())?;
-
     let output = child.wait_with_output()?;
     assert!(output.status.success());
 
--- a/docs/stdlib.dj
+++ b/docs/stdlib.dj
@@ -2264,35 +2264,21 @@ $ MYFLAG=42 jaq -n '$ENV.MYFLAG'
 
 The filter `env` is equivalent to `$ENV`.
 
-### `input_filename`, `$input_filename`
+### `input_filename`
 
-The filter `input_filename` returns:
+The filter `input_filename` returns
+the name of the file currently read, or
+`"<stdin>"` if standard input is currently read.
+If `--null-input` is used, this filter returns `null` instead.
 
-- The name of the file whose input is currently being filtered;
-- `"<stdin>"` when filtering data on standard input;
-- Or `null` if the data source is unknown (such as when using `--null-input`).
-
-This filter is implemented on top of the `$input_filename` variable, which gets set by the CLI automatically.
-
-Note that the CLI will only set the variable if it has not been explicitly set on the command line; for example,
-`jaq 'input_filename' ./data.json` will output `"./data.json"`, but
-`jaq --argjson input_filename 123 'input_filename' ./data.json` will output `123` instead.
-
-::: Compatibility
-`jq`'s version of the `input_filename/0` filter returns `"<stdin>"` when filtering data on standard input. While this
-behaviour has been preserved for `jaq`'s `input_filename/0` filter, the filter alone thus cannot be used to
-differentiate between filtering on standard input and filtering a file called `<stdin>`.
-
-To work around this, the `jaq` CLI sets the `$input_filename` variable to the actual input source: the path (if known),
-the _empty string_ for filtering data on standard input, or `null` if the data source is unknown. If your script needs
-to be able to distinguish between filtering data on standard input and filtering a file called `<stdin>`, use the
-variable rather than the filter.
-:::
-
-::: Advanced
-This filter and variable are only defined in the CLI, so are unavailable if you are using the `jaq-*` crates directly
-from Rust code.
-:::
+```
+$ jaq --slurp 'input_filename' "examples/benches.json"
+"examples/benches.json"
+$ echo 0 | jaq 'input_filename'
+"<stdin>"
+$ jaq -n 'input_filename'
+null
+```
 
 ### `repl`
 
--- a/docs/stdlib.dj
+++ b/docs/stdlib.dj
@@ -2264,6 +2264,36 @@ $ MYFLAG=42 jaq -n '$ENV.MYFLAG'
 
 The filter `env` is equivalent to `$ENV`.
 
+### `input_filename`, `$input_filename`
+
+The filter `input_filename` returns:
+
+- The name of the file whose input is currently being filtered;
+- `"<stdin>"` when filtering data on standard input;
+- Or `null` if the data source is unknown (such as when using `--null-input`).
+
+This filter is implemented on top of the `$input_filename` variable, which gets set by the CLI automatically.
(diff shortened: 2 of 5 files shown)


01mf02:
Hi @a-n-d-r-e-w-l, this is quite impressive work! Thanks!

A few concerns:

- I would not bake the `$jaq_input_filename` handling into `jaq-core`, especially not into the compiler. So far, there are no "magic" variables in there, and I would like to keep it like that.
- I'm not a fan of having `binds()` return an `Option<usize>` --- people that read the source will not understand what that does just by looking at the type signature.
- I'm not sure whether it's a good idea to expose `$jaq_input_filename` to the user and making it overrideable.
- Naming-wise, I did not introduce a single variable/filter starting with "jaq" so far, so I think I would rather like this to be just `$input_filename`.
- You write about files with filename `""` --- according to [this](https://unix.stackexchange.com/a/83793/675556), such files cannot exist. I would therefore just bind `$input_filename` to `null` if jaq is called with `-n`, to `""` if jaq reads from stdin, and to the file name otherwise.
- `input_filename($stdin)` is a nice idea in principle, but what if upstream jq decides to introduce that filter with different semantics? For that reason, I would stick to just `input_filename/0` for now and have it transform `""` to `"<stdin>"`. A user that really wants the distinction could just read `$input_filename`.

I would push `"input_filename"` to `var_names` and the correct corresponding value to `vars`, as well as move the definitions of `input_filename` into a new file `jaq/src/defs.jq`. I understand that this will make that functionality harder to use for consumers of `jaq-core`, but at least, they will get an error message à la `input_filename/0 is not defined`, rather than getting `null` all the time.

A lot of things to address here, I know, but I hope that I motivated my reasoning sufficiently. :) Thank you already so far for your great work, especially the documentation.

a-n-d-r-e-w-l:
> A lot of things to address here, I know, but I hope that I motivated my reasoning sufficiently.

Yep, that all makes sense! A lot of what you're saying I agree with and had thought about to some degree already (not liking magic variables, overriding it being a potential annoyance, etc.).

Of note (I should have checked in the preview not just the markdown editor) is that the return value in `jq` when filtering stdin is `"<stdin>"`, not `""` - I forgot to put it in a code block so it got interpreted as an HTML tag and stripped (oops); you're right about using `""` as a sentinel for stdin instead of `{}`, though.

The reason I went with `$jaq_input_filename` over just `$input_filename` is because it _is_ overridable, and the former is less likely to cause issues than the latter (it's fairly logical for someone's script to define a `$input_filename`, and far _less_ likely for them to add a `jaq_` prefix to it).

I'm not sure _how_ to approach making it non-overridable without introducing a breaking change of the form "you now cannot use `--arg input_filename foo`" - I'll poke around and see if I can come up with something.

tl;dr: I'll have a second pass at it and see if I can address all these points.

a-n-d-r-e-w-l:
Alright, pushed new changes:

- Magic variable handling in `jaq-core` removed
- Variable renamed to `$input_filename`
- Adding the variable now takes place in `real_main` rather than `binds`, so that the return value of the latter doesn't have an `(_, Option<usize>)` in it (this also keeps the entire `$input_filename`-handling logic within a single function for easier understanding at a glance)
- `input_filename/1` removed in favour of documenting the behaviour of `$input_filename` and when to use it over `input_filename/0`
- `input_filename/0` definition moved to new `jaq/src/defs.jq` to keep it within the CLI crate only

---

The one thing I don't have a solution for is the fact that the variable is overridable; it doesn't show in `$ARGS` unless set explicitly, so allowing it to be overridden means that this doesn't introduce any breaking changes that would prevent existing `jaq` invocations from working.

Note that **this is different** to how, say, `$ARGS` itself works - `jaq -cn --argjson ARGS 123 '$ARGS'` outputs `{"positional":[],"named":{"ARGS":123}}`, so the builtin variable setting takes precedence over the explicit argument being passed. Importantly, `$ARGS` is a variable present in original `jq` (which has the same builtin-set-takes-precedence as `jaq` does), but `$input_filename` is a variable new to `jaq` that doesn't come from `jq`.

Personally, I think that letting `$input_filename` be overridden is fine - it means no breaking changes for existing (valid) `jaq` invocations, no need to smuggle the value through something like an `Arc<Mutex<_>>`, no special-case handling for the variable name in `jaq-core`, etc. I've made sure that the documentation is explicit about this behaviour, so hopefully it won't end up as a surprise to anyone.

01mf02:
Ah, now I understand the problem with the `input_filename` overriding! If you use `jaq -n --argjson input_filename 123 '$input_filename'`, then this will not output `123`, but `null`! That's serious. Hmmm ... let me think a bit about how to handle this.

01mf02:
Just for reference, jq also overwrites user-defined variables with internal ones; i.e., `jq -n --argjson ENV 1 '$ENV'` yields `{...}`, not `1`.

01mf02:
> Personally, I think that letting `$input_filename` be overridden is fine - it means no breaking changes for existing (valid) `jaq` invocations

If you run `jq -n --arg input_filename 1 '$input_filename'`, you would get 1. Same with jaq 3.0. But with the current PR, you would get `null`, so this looks like a breaking change to me?

01mf02:
One option would be to bind the `input_filename` to a variable such as `$!input_filename` (note the `!`), which cannot be referenced by a user. However, we then also cannot implement `input_filename` by definition, because we also cannot reference `$!input_filename` in a regular definition. We could, however, write Rust code that produces a filter that *can* reference `$!input_filename`, similar to the code that jaq uses to generate a definition [`!empty`](https://github.com/01mf02/jaq/blob/961601926ac4c8e8281a817aa21e20a1ee53b04c/jaq-core/src/load/parse.rs#L184). This might be a bit verbose, however.

01mf02:
Let me play around with this idea a bit, and please do not commit anything in the meanwhile.

jaq/tests/golden.rs
```diff
@@ -18,7 +19,10 @@ fn golden_test(args: &[&str], input: &str, out_ex: &str) -> io::Result<()> {
     if out_ex.trim() != out_act.trim() {
         println!("Expected output:\n{}\n---", out_ex);
         println!("Actual output:\n{}\n---", out_act);
-        process::exit(2);
+        return Err(io::Error::new(
```
01mf02, line 21: Is there a reason for this change? Or can I revert it back?

01mf02:
@a-n-d-r-e-w-l, it seems that my `$!input_filename` approach was not that verbose after all. Unfortunately, as side effect, the user now cannot distinguish a file `<stdin>` from actual `stdin`, but given that this is precisely the behaviour that jq implements too, I think that I can live with that.

I removed some of the documentation that you added for the API, because the HTML docs are not intended to cover the API. Otherwise, I would need to add many, many more things there, making it more confusing for users that just want to use jaq as CLI. I hope that you understand.

I'm waiting for your feedback on the golden tests issue, and after that I think that this looks ready to merge! :)

jaq/tests/golden.rs
a-n-d-r-e-w-l: IIRC all tests get run concurrently _within the same process_, so using `std::process::exit` (or `::abort`) will kill the test harness at the first failure, while returning an error means that multiple failures can be reported at once.

a-n-d-r-e-w-l:
> If you run `jq -n --arg input_filename 1 '$input_filename'`, you would get 1. Same with jaq 3.0. But with the current PR, you would get `null`, so this looks like a breaking change to me?

I don't think so?
```fish
$ for command in "cargo run -q --" "jaq" "jq"; c=(string split ' ' $command) $c -n --argjson input_filename 1 '$input_filename'; end
1
1
1
```

> Just for reference, jq also overwrites user-defined variables with internal ones; i.e., `jq -n --argjson ENV 1 '$ENV'` yields `{...}`, not `1`.

Yeah, the reason that I treated `$input_filename` differently to that is because it's not a variable that jq itself uses, so silently overriding it can cause existing scripts that jaq 3.0.0 can run to suddenly diverge from what they used to do, and from what jq does.
What I'm talking about is some existing script that is invoked as `jq -n --arg input_filename foo 'some_filter | [., $input_filename]'` (unlikely perhaps, but valid). With 3.0.0, both jaq and jq agree what the output is (`["result of some_filter", "foo"]`). With this PR - where the variable does _not_ get overwritten by the CLI - they both still agree. If `$input_filename` is treated like `$ENV` and _always_ get overwritten by the CLI, then the output would change to `["result of some_filter", "/path/to/input/file"]` - there hasn't been an error or warning, but there's been a silent change of behaviour. This is technically fine over a major version bump, but probably still undesirable.

> Unfortunately, as side effect, the user now cannot distinguish a file `<stdin>` from actual `stdin`

That is mildly annoying, but any ability to do so would only be a slight improvement on upstream so isn't actually required - also, it's one of the most unlikely usages I can think of so probably won't happen regardless. (writing a JSON file _called_ `<stdin>` seems unusual to say the least, and the command line would need to be `jaq 'input_filename' "<stdin>"` without a leading `./` on the path, so even more unlikely still)

---

I've left one comment about using `.to_string_lossy()` rather than `.as_encoded_bytes()`, but other than that your changes look good - I had considered using a variable name that wouldn't parse, but a quick search didn't make any existing usages of that pattern jump out at me, so I figured it may not work for some reason.

jaq/tests/golden.rs
01mf02: I see. Then this is a good thing to keep. Thanks for the clarification!

01mf02:
> I don't think so?
> 
> ```fish
> $ for command in "cargo run -q --" "jaq" "jq"; c=(string split ' ' $command) $c -n --argjson input_filename 1 '$input_filename'; end
> 1
> 1
> 1
> ```

Ah yes, you are right. That's because you originally checked whether the user has set `input_filename`, and do not overwrite it in that case. However, that means that the confusing behaviour is then the following:

~~~
# your latest commit on this branch
$ git checkout 7e578dce4923734175b6f8e1156d52d08a689d4d
$ cargo run -q -- -n --arg input_filename 1 'input_filename'
"1"
$ jq -n --arg input_filename 1 'input_filename'
null
~~~

But I think that you already pointed this out before above somewhere.

01mf02:
> I've left one comment about using `.to_string_lossy()` rather than `.as_encoded_bytes()`, but other than that your changes look good - I had considered using a variable name that wouldn't parse, but a quick search didn't make any existing usages of that pattern jump out at me, so I figured it may not work for some reason.

I'm sorry, I do not find that comment --- can you point me to it?

In any case, I would have preferred to use `.as_encoded_bytes()` the way you did, but that would imply yet another MSRV and thus longer CI build times. Unless there is another way to get the underlying bytes of an [`OsStr`](https://doc.rust-lang.org/std/ffi/struct.OsStr.html) in Rust 1.70?

a-n-d-r-e-w-l:
In terms of the "CLI not setting `$input_filename` if an explicit variable of that name exists", I figured it would be the better option than "always set the variable" because then at least existing scripts that _currently work in jaq_ wouldn't change behaviour - though, of course, using `$!input_filename` instead from your changes fixes this.

(note for future readers who have the same thought I do: `jaq --arg '!input_filename' foo '$ARGS.named["!input_filename"]'` **won't** be affected by the CLI's setting of `$!input_filename`, as `$ARGS` gets built before the variable gets overwritten; given that `$ARGS` is the only way to read variables whose names are invalid identifiers, this means that there is no filter-accessible way to observe the fact that the CLI uses this variable to implement the filter)

---

> I'm sorry, I do not find that comment --- can you point me to it?

I probably should have clarified - I meant PR review comment [[here](https://github.com/01mf02/jaq/pull/432/changes#r3190231316)], not `// Rust comment`.

Damn, I completely forgot about MSRV. No, there unfortunately doesn't seem to be a way to get the underlying bytes of an `OsStr` in 1.70 (well, you _can_ on WASI and Unix, but that's not cross-platform enough).

If sticking on 1.70, `.to_string_lossy()` does seem to be the only option - though it's then probably worth adding a `FIXME` or similar comment so that it can be fixed once MSRV is bumped.

---

It is worth noting that 1.70 was released on 2023-06-01, making it pretty much three years old; 1.74 was released on 2023-11-16, so is ~2.5 years old. (also, _just_ the `jaq-core` crate is on 1.69 [2023-04-20] instead of 1.70, which may not be intentional)

Given that this particular missing function results in a correctness failure (`.to_string_lossy()` means that the _actual_ filepath does not get output, just an approximation of it), I'd argue that it's probably worth a bump to MSRV.

If the 1.70 MSRV is to allow "old" Rust versions to still compile jaq, I'd say that 1.74 is probably _also_ "old" enough to be fine to bump to.

If the only reason to keep the 1.70 MSRV is CI times (so not support for nearly 3 year old versions of Rust), then bumping to 1.85 [2025-02-20, just over a year old] would be a good choice _anyway_; the [perf comparison site](https://perf.rust-lang.org/compare.html?start=2023-05-31&end=2025-02-17&stat=wall-time) claims an average ~16% compile-time _reduction_ between 1.70 and 1.85. Some testing on my machine shows that `cargo clean; time cargo +$VER build --release --offline` goes from an average of ~22.3s on 1.70 to ~19.1s on 1.85 (with prewarmed SSD cache, three trials each) - roughly 14% reduction, pretty close to the perf report average. The same but for debug builds goes from 6.0s to 5.2s, a 13% saving [^1]. There would also be runtime performance and memory usage improvements, but it's harder to get concrete numbers for those. [^2]

It's your call if you want an MSRV bump at all; in my opinion, the 1.74 bump at a minimum would make sense so that the output of this can be correct, while a bump to 1.85 (or higher) would be better.
- If you don't want a bump, I'll just add a comment to the source to remind Future Contributor to swap it out for `.as_encoded_bytes()` when the MSRV _is_ bumped.
- If you _do_ want a bump, I'm happy to do the MSRV[+edition bump]/lint fixes/formatting/etc. in a separate PR - just let me know what version to bump to

It may also be worth documenting what the MSRV decision is - the one I've seen most commonly across projects that enforce MSRV is "stable minus 2" (i.e. about three months), but the amount of leeway does vary somewhat - so that future MSRV bumps can happen somewhat more regularly.

[^1]: Just for fun, testing 1.95 (i.e. current stable):  
The same three-trial-with-cleaning setup results in 18.8s average for release builds (1.8% better than 1.85, 15.9% better than 1.70), and 5.0s average for debug builds (4.2% better than 1.85, 16.8% better than 1.70).
[^2]: It's also worth noting that the size of the resulting binary goes from 4.55 MB to 4.34 MB for release mode (4.4% savings), and from 117.1 MB to 74.9 MB for debug mode (_36%_ savings).

01mf02:
> Given that this particular missing function results in a correctness failure (`.to_string_lossy()` means that the _actual_ filepath does not get output, just an approximation of it), I'd argue that it's probably worth a bump to MSRV.

After reflection, I would be for bumping MSRV, but only if the [`OsStr::as_encoded_bytes idea`](https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.as_encoded_bytes) is the right thing to use here. However, I have my doubts after reading the documentation:

> The byte encoding is an unspecified, platform-specific, self-synchronizing superset of UTF-8. By being a self-synchronizing superset of UTF-8, this encoding is also a superset of 7-bit ASCII.
> Note: As the encoding is unspecified, any sub-slice of bytes that is not valid UTF-8 should be treated as opaque and only comparable within the same Rust version built for the same target platform. For example, sending the slice over the network or storing it in a file will likely result in incompatible byte slices. See [OsString](https://doc.rust-lang.org/std/ffi/struct.OsString.html) for more encoding details and [std::ffi](https://doc.rust-lang.org/std/ffi/index.html) for platform-specific, specified conversions.

I'm worried about something like `jaq 'input_filename' file_with_invalid_utf8_chars_in_filename.json`. If I understand the docs above correctly, this may then yield different outputs depending even the Rust version and the target platform. Or how do interpret this?

01mf02:
> It may also be worth documenting what the MSRV decision is - the one I've seen most commonly across projects that enforce MSRV is "stable minus 2" (i.e. about three months), but the amount of leeway does vary somewhat - so that future MSRV bumps can happen somewhat more regularly.

Where would you expect this documentation to be?

My decision on MSRV is to always use the smallest one possible that allows the implementation of all desired features.

01mf02:
About the CI performance: I test in CI for every package in this repository whether it compiles with its specified MSRV. So having three different MSRVs takes longer for CI than having just two. Because my policy is to always use the smallest MSRV possible, I cannot not just increase MSRV to decrease CI build times. :)

a-n-d-r-e-w-l:
> If I understand the docs above correctly, this may then yield different outputs depending even the Rust version and the target platform. Or how do interpret this?

My understanding of this is as follows:
- Any subslice of `.as_encoded_bytes()` that _is_ valid UTF-8 is, well, valid UTF-8 and can thus be sent over the network or stored in a file or whatever
  - Notably, this means that for the (these days) overwhelmingly common case of fully-UTF-8 filepaths then they can just be treated as normal text
- Any non-UTF-8 return value represents the path _on that system_, and can safely be round-tripped back through `OsStr::from_encoded_bytes_unchecked` _on that same system_; the reason that bytes from a file or the network cannot be put back through this is that you can't be safely certain that they originate from a `.as_encoded_bytes()` on a system that represents paths in **exactly** the same manner.

---

> Where would you expect this documentation to be?

Generally (from what I've seen) it tends to be a short block at the end of the crate-level documentation.

> My decision on MSRV is to always use the smallest one possible that allows the implementation of all desired features.

Yep, that makes sense! For that reasoning, a bump to 1.85 or higher wouldn't fit, so just the possible bump to 1.74 is left.

---

I've also just done a bit more digging to see how jq handles this itself.

To cut a long story <sup>[[filter registration](https://github.com/jqlang/jq/blob/jq-1.8.1/src/builtin.c#L1959); [filter implementation](https://github.com/jqlang/jq/blob/jq-1.8.1/src/builtin.c#L1857); [util implementation](https://github.com/jqlang/jq/blob/jq-1.8.1/src/util.c#L379); [setting the state during parsing](https://github.com/jqlang/jq/blob/jq-1.8.1/src/util.c#L284); string constructor definition [1](https://github.com/jqlang/jq/blob/jq-1.8.1/src/jv.c#L1290), [2](https://github.com/jqlang/jq/blob/jq-1.8.1/src/jv.c#L1280-L1282), [3](https://github.com/jqlang/jq/blob/jq-1.8.1/src/jv.c#L1112)]</sup> short, it turns out that jq actually does the equivalent of `.to_string_lossy()` for filenames - so while `.as_encoded_bytes()` may be more correct from the filesystem perspective, `.to_string_lossy()` is actually correct from an _upstream compatibility_ perspective.

I probably should have just checked upstream earlier, but the end result is that your usage of `.to_string_lossy()` is most likely the right one for this filter - so no MSRV bump is actually needed in the end.

01mf02:
> To cut a long story [[filter registration](https://github.com/jqlang/jq/blob/jq-1.8.1/src/builtin.c#L1959); [filter implementation](https://github.com/jqlang/jq/blob/jq-1.8.1/src/builtin.c#L1857); [util implementation](https://github.com/jqlang/jq/blob/jq-1.8.1/src/util.c#L379); [setting the state during parsing](https://github.com/jqlang/jq/blob/jq-1.8.1/src/util.c#L284); string constructor definition [1](https://github.com/jqlang/jq/blob/jq-1.8.1/src/jv.c#L1290), [2](https://github.com/jqlang/jq/blob/jq-1.8.1/src/jv.c#L1280-L1282), [3](https://github.com/jqlang/jq/blob/jq-1.8.1/src/jv.c#L1112)] short, it turns out that jq actually does the equivalent of `.to_string_lossy()` for filenames - so while `.as_encoded_bytes()` may be more correct from the filesystem perspective, `.to_string_lossy()` is actually correct from an _upstream compatibility_ perspective.
> 
> I probably should have just checked upstream earlier, but the end result is that your usage of `.to_string_lossy()` is most likely the right one for this filter - so no MSRV bump is actually needed in the end.

Then let's keep things like they are right now. Thanks for all your research!

01mf02:
Uh-oh, bad news ahead:

~~~
$ jq -ns '(., inputs) | input_filename' examples/benches.json 
null
"examples/benches.json"
$ cargo run -q -- -ns '(., inputs) | input_filename' examples/benches.json 
null
null
~~~

~~~
$ echo 1 | jq -n '(., inputs) | input_filename'
null
"<stdin>"
$ echo 1 | cargo run -q -- -n '(., inputs) | input_filename'
null
null
~~~

My intuition would be to simply set `$!input_filename` even if `--null-input` is set. The jq manual does not actually specify what happens in the case of `--null-input`, so ... :shrug:
What do you think about that?

01mf02:
I have now made the change to set `input_filename` even if `-n` is given.

01mf02:
Thanks again for all your feedback, @a-n-d-r-e-w-l!

PR #442, opened 2026-06-15
Convert manual to man page without Pandoc

01mf02:
This generates the man page with a little jq script, thus removing the dependency on Pandoc.

Note that this removes "advanced" blocks, because these contain tables, which already looked broken in Pandoc's output.


01mf02:
@alerque, sorry to bother you, but I tried to contact you several times by mail about including the man page in the Arch Linux package of jaq. Because I never received an answer from you, I'm not sure that my mail got through to you. In any case, this PR would make man page building significantly easier on Arch Linux. If I merge it and release a new (minor) version of jaq, would you be willing to include the man page in jaq's Arch Linux package? I could also send you a diff for the PKGBUILD, if you wish.

alerque:
Sorry email is not a good way to find me, especially cold-turkey or about packaging things which get spammed like crazy. I probably have 100,000 unread messages related to packaging and 99.999% of them are automated cruft. It's hard to spot anything sensible coming through--don't take it personal! The best thing would be an issue (or ever MR) on Arch's packaging bug tracker.

Of course we can package the man page ... whether hard to build or not. Pandoc is bulky but as just a build time dependency it isn't the end of the world either.

01mf02:
@alerque, thanks for your reply, and thanks for explaining the email issue.

In that case, I suggest to wait for jaq 3.1.1 to be released with Pandoc-free manpage building, and then update the PKGBUILD. I tried to register for an Arch Linux account, but I failed to login both via GitHub and registering a new account failed because my Recaptcha answer was not accepted (output of `LC_ALL=C pacman -V|tail -n3|base32|head -1`).

Should I ping you about this once I have released jaq 3.1.1?

alerque:
https://gitlab.archlinux.org/archlinux/packaging/packages/jaq/-/commit/927a7e48471cb7fb3098f17ad037934f15909ae0

If the make command fails on the next release I'll know why and come looking for the correct invocation.

docs/man.jq
```diff
@@ -0,0 +1,41 @@
+# Convert HTML to man page.
```
alerque, line 1: For *any* other project I'd say this was an absurd way to work around not wanting to use a standard tool. I guess for this project to dog food itself is ... chaotic genius‽

01mf02: I feel genuinely honoured by your remark.

docs/Makefile
```diff
@@ -4,16 +4,22 @@ FILES=intro.dj cli.dj corelang.dj stdlib.dj advanced.dj formats.dj examples.dj
...
+%.xhtml: %.dj
+	(echo "<body>" && jotdown $< && echo "</body>") > $@
+
 MANUAL.xhtml: manual.jq body.xhtml style.css template.xhtml
 	$(JAQ) -f manual.jq --rawfile body body.xhtml --rawfile style style.css template.xhtml --to xml > $@
 
-jaq.1: filter.lua man-prologue.dj $(FILES) man-epilogue.dj
-	pandoc -V title=JAQ -V section=1 --lua-filter filter.lua --from djot man-prologue.dj $(FILES) man-epilogue.dj -s -o $@
+%.man: %.xhtml man.jq
+	$(JAQ) -f man.jq $< -j > $@
```
oech3, line 19: Can I omit GNU `make` too by the way?

01mf02: To build the man page? If you are hoping for me to reimplement GNU make in jaq, then I'll unfortunately have to disappoint you. :)

But if you are fine to write down the commands that `make` executes to generate the man page, sure thing, why not, then you don't need GNU make. I just will not guarantee that the commands will not change, so if I were you, I'd stick to GNU make.

oech3: Ok. I'll try copy and pasting the command. Thankyou.
`make` is dep if `base-devel`, but I prefer less dep.

alerque: @oech3 You're crossing over into crazy land now. Do not avoid invoking `make` just to avoid the dependency in an Arch Linux package build. Building AUR packages *ABSOLUTELY REQUIRES* base-devel as a minimum starting point, and `make` is included in that.

Valid reasons for not using an upstream provided make rule would be in the Makefile did something that went against platform specific packaging guidelines. This is note one of those cases.

oech3: I just used `make` and `jotdown` since it seems complex more than I considered.

Issue #444, opened 2026-06-27
-help | : Broken Pipe should not be shown


oech3:
```
$ jaq --help|:
Error: Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }
$ jq --help|: # silent
```

PR #445, opened 2026-06-29
Handle broken pipes more gracefully

01mf02:
This handles broken pipes as reported in #444:

~~~
$ cargo run -q -- --help | :
$ echo "${PIPESTATUS}"
141
$ cargo run -q -- -n 1 | :
$ echo "${PIPESTATUS}"
141
~~~

It would be nice to unify at least two of the three broken pipe checks in the code, though ...
But this is hard, given that error printing code can fail with `SIGPIPE` as well ...

### DIFF
--- a/jaq-fmts/src/write/mod.rs
+++ b/jaq-fmts/src/write/mod.rs
@@ -33,8 +33,7 @@ pub struct Writer {
 }
 
 /// Buffer writes if stdout is terminal, else just lock stdout.
-pub fn with_stdout<T>(f: impl FnOnce(&mut dyn Write) -> T) -> T {
-    let stdout = io::stdout();
+pub fn with_stdout<T>(stdout: &mut io::Stdout, f: impl FnOnce(&mut dyn Write) -> T) -> T {
     if stdout.is_terminal() {
         f(&mut stdout.lock())
     } else {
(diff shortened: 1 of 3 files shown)


oech3:
`unsafe { libc::signal(libc::SIGPIPE, libc::SIG_DFL) };`? Taken from uutils.

01mf02:
> `unsafe { libc::signal(libc::SIGPIPE, libc::SIG_DFL) };`? Taken from uutils.

I rather wish to avoid unsafe code, as well as adding another (direct) dependency to `jaq`.

oech3:
I don't think thread-safe wrapper for signal is possible. (It should be used by uutils already if such wrapper is possible...)

01mf02:
> I don't think thread-safe wrapper for signal is possible. (It should be used by uutils already if such wrapper is possible...)

Perhaps, but jaq does not use multiple threads, so do you think that this is relevant in this case?

oech3:
External crates or C libs might use threads. I'm not sure, but is signal on top of main() safe?

01mf02:
> External crates or C libs might use threads. I'm not sure, but is signal on top of main() safe?

To be honest, I do not know. Problematic places could be filters like `debug`, `stderr`, which are printed via `env_logger`, and I do not know how that works under the hood. If you find some way to trigger unexpected behaviour there, I would be interested to hear about it.

PR #448, opened 2026-07-15
Disable default features for workspace dependencies

### LINKED ISSUE #431: jaq-all pulls in all jaq-fmts formats
Really enjoying working with jaq!

I'm trying to use the batteries-included `jaq-all` crate, but noticed that, despite putting `jaq-fmts/all` behind a Cargo feature, it pulls in all format dependencies even with `default-features = false`.

I _think_ this is because `jaq-all` uses the workspace `Cargo.toml` declaration of the `jaq-fmts` dependency, which pulls in all features, even if they're not requested by `jaq-all` itself.

It looks like adding `default-features = false` in the workspace `Cargo.toml` would let the Cargo features in `jaq-all` take effect, but not sure if I'm missing something!

### LINKED ISSUE #446: Dependency structure pulls in transitive dependencies with all default features
I’m trying to reduce my dependency count, but `jaq` makes this virtually impossible. The primary problem is that the `workspace = true` dependencies always enable all features on those dependencies (e.g. `jaq-std` pulled in by several other crates). Some dependencies like `jiff` have tons of useless default dependencies for many use cases, such as `defmt`. However, disabling the `time` feature on `jaq-std` has no effect, since `jaq-fmts` pulls in `jaq-std` with default features anyways. So I can’t get rid of `defmt` in my dependency tree, despite `jaq` not needing it and me not needing it when interacting with `jiff`.

This is related to #431, but aimed at the entire workspace tree - please set `default-features = false` everywhere and only enable the features required by `jaq`.

kleinesfilmroellchen:
This allows actually disabling features on the various libraries and that having an actual effect. Now, by default, dependent features are not enabled unless the user asks for them or the library absolutely requires it. For example, jaq-all by default requires pretty much all features on jaq-std but this can be disabled by disabling the (only) new feature all-std.

This also adds the correct config for the tabular feature in jaq-fmts, which wouldn’t compile with this feature disabled beforehand.

This appears to be a breaking change for jaq-all if you disable default features (not if you leave them enabled), since jaq-all is missing three functions without the new "all-std" feature. However it’s not breaking for jaq itself (which still enables everything), or jaq-json (which just has to do some more work under the `std` feature), or jaq-core (which is untouched). For jaq-fmts this is a bugfix, enabling compilation with the tabular feature disabled.

Fixes #446. Fixes #431, where you are required to disable the new `all-std` feature, which also disables a few functions.

I have verified the dependency tree under --no-default-features in the relevant places and confirmed that the dependencies are minimal and as I expect them.

### Checklist

- [x] I have read the [contribution guidelines](../CONTRIBUTING.md).

### DIFF
--- a/jaq-fmts/src/write/mod.rs
+++ b/jaq-fmts/src/write/mod.rs
@@ -1,6 +1,7 @@
 //! Write values in different formats.
 #[cfg(feature = "cbor")]
 pub mod cbor;
+#[cfg(feature = "tabular")]
 pub mod tabular;
 #[cfg(feature = "toml")]
 pub mod toml;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,10 +13,10 @@ resolver = "2"
 
 [workspace.dependencies]
 jaq-core = { version = "3.1.0", path = "jaq-core", default-features = false }
-jaq-std  = { version = "3.0.1", path = "jaq-std", default-features = false  }
+jaq-std  = { version = "3.0.1", path = "jaq-std" , default-features = false }
 jaq-json = { version = "2.0.1", path = "jaq-json", default-features = false }
 jaq-fmts = { version = "0.1.0", path = "jaq-fmts", default-features = false }
-jaq-all  = { version = "0.2.0", path = "jaq-all", default-features = false  }
+jaq-all  = { version = "0.2.0", path = "jaq-all" , default-features = false }
 
 [profile.release]
 strip = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,11 +12,11 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-jaq-core = { version = "3.1.0", path = "jaq-core" }
-jaq-std  = { version = "3.0.1", path = "jaq-std"  }
-jaq-json = { version = "2.0.1", path = "jaq-json" }
-jaq-fmts = { version = "0.1.0", path = "jaq-fmts" }
-jaq-all  = { version = "0.2.0", path = "jaq-all"  }
+jaq-core = { version = "3.1.0", path = "jaq-core", default-features = false }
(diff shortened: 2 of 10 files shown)


01mf02:
Hi @kleinesfilmroellchen! Thanks a lot for this PR. I will have a more thorough look later. I'm a bit concerned about the `all-std` approach, but I'll think about it.

kleinesfilmroellchen:
Unfortunately having these things under a feature (possibly an existing one) is the only real solution, and the only one that addresses my usecase.

01mf02:
> Unfortunately having these things under a feature (possibly an existing one) is the only real solution, and the only one that addresses my usecase.

OK, I see. I think that `std-all` suits me better as feature name, because it reflects writing `jaq-std/all`. Note that it now actually activates `jaq-std/default`, which is a simpler way than writing `jaq-std/regex`, `jaq-std/log`, ...

What do you think about my changes and the overall state?

jaq-fmts/src/write/mod.rs
```diff
@@ -1,6 +1,7 @@
 //! Write values in different formats.
 #[cfg(feature = "cbor")]
 pub mod cbor;
+#[cfg(feature = "tabular")]
```
01mf02, line 4: Thanks for spotting this!

kleinesfilmroellchen:
> > Unfortunately having these things under a feature (possibly an existing one) is the only real solution, and the only one that addresses my usecase.
> 
> OK, I see. I think that `std-all` suits me better as feature name, because it reflects writing `jaq-std/all`. Note that it now actually activates `jaq-std/default`, which is a simpler way than writing `jaq-std/regex`, `jaq-std/log`, ...
> 
> What do you think about my changes and the overall state?

I like it. I wasn't sure about the warnings, so I left them alone originally. I agree with the name, no strong feelings. If you're OK with everything else you can merge the PR :)

01mf02:
Great --- thanks again for your contribution! :)

PR #449, opened 2026-07-18
Flush output per value only when stdout is a terminal

vlasky:
Previously, `write()` flushed after every output value, which defeated the `BufWriter` that `with_stdout()` sets up for non-terminal output: each value cost one write syscall. On `jaq -n 'range(200000)' > /dev/null`, roughly 60% of the runtime was system time spent on these writes.

This PR moves the per-value flush out of `write()` and into the callers:

- `with_stdout()` now tells its closure whether stdout is a terminal, and flushes the buffer once at the end otherwise. It also propagates flush errors, which `BufWriter`'s silent flush-on-drop would previously have swallowed, and it flushes even when the closure fails, so that a filter error cannot mask a stdout failure.
- The CLI flushes per value only when stdout is a terminal, which preserves the behaviour that makes `"prompt> "` appear before `input` blocks in `jaq -jn '"prompt> " | (., input)'`.
- A new `--unbuffered` flag (matching jq) forces a flush after every output value even when stdout is not a terminal, for protocol-style usage such as a parent process reading jaq's output through a pipe. Documented in `docs/cli.dj`.
- The REPL always flushes per value, since it is interactive by nature.
- The `jaq-all` example flushes explicitly, since `write()` no longer does.

One known limitation, documented on `with_stdout()`: nested calls buffer independently, so output of a `repl` invoked with redirected stdout can appear before buffered output of the outer run. `--unbuffered` restores the previous ordering.

```
jaq -n 'range(200000)' > /dev/null    96.5 ms -> 16.6 ms  (5.8x)
```

Output is byte-identical to the previous behaviour in all cases; the change only affects when bytes reach the operating system.

Note that the changed `with_stdout()` signature is a source-breaking change for `jaq-fmts` users: the closure gains an `is_terminal` argument and must now return a `Result` whose error implements `From<io::Error>`.

Testing: full workspace test suite, the docs pipeline (`make all` in `docs/`, including the 89 embedded CLI tests), `cargo clippy -- -Dwarnings`, `cargo fmt --check`, and MSRV checks with Rust 1.69 (jaq-core) and 1.70 (workspace) all pass.

### DIFF
--- a/docs/cli.dj
+++ b/docs/cli.dj
@@ -354,6 +354,21 @@ $ echo [1, [2]] | jaq       | jaq -Rs
 
 Use _N_ spaces for indentation (default: 2).
 
+{#--unbuffered}
+### `--unbuffered`
+
+Flush the output after each value.
+
+When the output is not a terminal, jaq buffers it and
+flushes it only once all values have been written.
+This flag makes jaq flush the output after each value instead,
+which is useful when another process consumes jaq's output as
+soon as it is produced; for example:
+
+```
+jaq --unbuffered .a live.json | while read -r a; do echo "got $a"; done
+```
+
 ### `NO_COLOR`
 
 If the environment variable `NO_COLOR` is set to a non-empty value, then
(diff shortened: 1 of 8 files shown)


01mf02:

PR #450, opened 2026-07-18
Replace per-call Aho-Corasick automata by simple byte loops

vlasky:
`escape_html`/`unescape_html` (jaq-std), `@csv`/`@tsv` field escaping (jaq-fmts), and the playground's HTML escaping (jaq-play) constructed a fresh Aho-Corasick automaton for every processed value. For short strings, the automaton build cost dwarfs the actual scan.

All of these replacements concern single bytes (or, for `unescape_html`, five fixed entities, none of which is a prefix of another), so a plain byte loop is equivalent to the previous `MatchKind::Standard` `replace_all` behaviour. For CSV/TSV, the loop also writes directly to the output instead of allocating an escaped copy of each field.

This removes the `aho-corasick` dependency from jaq-std, jaq-fmts, and jaq-play.

Benchmarks (Apple Silicon, release build):

```
map(@html) over 200k strings     2.36 s -> 81 ms  (29x)
.[] | @tsv over 50k rows         2.24 s -> 56 ms  (40x)
```

The second commit adds regression tests that pin down the preserved semantics: single-pass entity unescaping without rescanning (`"&amp;lt;"` yields `"&lt;"`), preservation of unknown or incomplete entities (`"&x;"`, `"&amp"`), and the escaping of control characters, backslashes, and quotes in TSV/CSV fields.

Verified byte-identical output against the previous build on an edge-case corpus (adjacent entities, lone ampersands, Unicode, embedded tabs/newlines/NUL), plus round-trip `escape_html | unescape_html` identity.

Testing: full workspace test suite, `cargo clippy -- -Dwarnings`, `cargo fmt --check`, MSRV checks with Rust 1.69 (jaq-core) and 1.70 (workspace), and `cargo check -p jaq-play --target wasm32-unknown-unknown` all pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 3
 
 [[package]]
-name = "aho-corasick"
-version = "1.1.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
-dependencies = [
- "memchr",
-]
-
-[[package]]
 name = "arbitrary"
 version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


01mf02:

PR #451, opened 2026-07-18
Cache the parsed value of decimal numbers

vlasky:
Non-integer numbers read from JSON/YAML input are stored as decimal literals to preserve their exact representation on output. Previously, every comparison, hash, or arithmetic operation on such a number re-parsed its literal via `from_dec_str`, so sorting or deduplicating float-heavy data paid one string parse per comparison operand.

`Num::Dec` now stores a `Dec` struct that couples the literal with its parsed `f64` (NaN for invalid literals, as before), so all operations read the cached value. Round-trip printing is unchanged: output still uses only the literal.

Two side effects that seem like improvements:

- The `Dec` fields are private, which guarantees the previously implicit invariant that the cached value matches the literal. (Previously, `Num`'s `Rc<String>` payload could be constructed with arbitrary contents by downstream code.)
- `Num::as_f64` no longer panics on literals that fail to parse; it returns NaN, consistent with `from_dec_str`.

This is a source-breaking change for code that matches on or constructs `Num::Dec` directly; `Num::dec(String)` is the new constructor, and `Dec` (re-exported from the crate root) provides `as_str`/`as_f64` accessors.

```
sort of 200k floats    155 ms -> 59 ms  (2.6x)
```

Verified byte-identical output against the previous build for decimal round-trips (`100.00`, `1.5e3`, `-0.0`, big-integer literals), `sort`, `unique`, and `group_by` over the float corpus.

Testing: full workspace test suite, `cargo clippy -- -Dwarnings`, `cargo fmt --check`, and MSRV checks with Rust 1.69 (jaq-core) and 1.70 (workspace) all pass.

### DIFF
--- a/jaq-fmts/src/read/yaml.rs
+++ b/jaq-fmts/src/read/yaml.rs
@@ -218,7 +218,7 @@ fn parse_float(s: &str) -> Option<Val> {
         let f = f64::INFINITY;
         Some(Val::from(if sign == Some('-') { -f } else { f }))
     } else {
-        normalise_float(sign, rest).map(|f| Val::Num(Num::Dec(f.into())))
+        normalise_float(sign, rest).map(|f| Val::Num(Num::dec(f)))
     }
 }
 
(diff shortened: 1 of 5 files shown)


01mf02:

PR #452, opened 2026-07-18
Parse decimal program literals at compile time

vlasky:
Builds on #451 (the first commit here is that PR's commit); please review/merge that one first, after which this PR reduces to its second commit.

Previously, a decimal literal such as `1.5` in a program was stored as a string and re-parsed at every evaluation: first as `isize`, then as big integer, then as `f64`. So `map(. * 1.5)` parsed `"1.5"` once per array element.

Now the compiler stores the parsed `f64` alongside the literal for decimal (non-integer) literals, and the new `ValT::from_dec(lit, num)` method lets jaq-json construct the decimal value directly. The method has a default implementation that falls back to `from_num`, so other value implementations are unaffected. Integer literals that overflow `isize` (which, by the lexer's number grammar, are the only non-decimal literals that reach this path) keep using `from_num`, preserving big integer semantics.

```
map(. * 1.5) over 500k ints    102 ms -> 89 ms
```

The remaining gap to integer constants is the allocation of the literal per produced value, which would need per-run value caching to remove.

Verified identical behaviour against the previous build for literal round-trips (`100.00`, `1.5e3`), big-integer program literals, `0.1 + 0.2`, `1.5 == 1.500`, and out-of-range literals (`1e309`).

Testing: full workspace test suite, `cargo clippy -- -Dwarnings`, `cargo fmt --check`, and MSRV checks with Rust 1.69 (jaq-core) and 1.70 (workspace) all pass.

### DIFF
--- a/jaq-core/src/compile.rs
+++ b/jaq-core/src/compile.rs
@@ -92,7 +92,10 @@ pub(crate) enum Term<T = TermId> {
     ToString,
 
     Int(isize),
-    Num(String),
+    /// Number literal that is not representable by [`Term::Int`],
+    /// optionally with its pre-parsed floating-point value
+    /// (present iff the literal is decimal, i.e. not an integer)
+    Num(String, Option<f64>),
     Str(String),
     /// Array construction (`[f]`)
     Arr(T),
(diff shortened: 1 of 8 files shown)


01mf02:

PR #453, opened 2026-07-18
Fix three arithmetic operations that abort or panic on overflow

vlasky:
Three arithmetic edge cases currently abort or panic the process on ordinary filter input. Each is fixed here with a regression test; the fixes are independent commits.

**1. String repetition (`jaq-json`).** `str * n` for a large `n` called `[u8]::repeat`, which aborts the process: `"ab" * 9223372036854775807` aborts with a capacity overflow (the length exceeds `isize::MAX`), and `"a" * 9223372036854775807` aborts with an allocation failure (the length is representable but too large to allocate). Both now bound the length and allocate fallibly with `try_reserve_exact`, returning jq's error `Repeat string result too long`. The non-overflowing path is unchanged in behaviour and stays O(log n) in copies via `extend_from_within`.

**2. `length` of `isize::MIN` (`jaq-json`).** `Num::length` used `i.abs()`, which panics in debug and wraps to a negative length in release for `isize::MIN`. It now promotes to a big integer in that case, mirroring `Num::neg`.

**3. Epoch to timestamp (`jaq-std`).** `epoch_to_timestamp` multiplied the epoch seconds by 1_000_000 without checking for overflow, so a large integer epoch (e.g. `100000000000000000 | gmtime`, also `strftime`/`localtime`) panicked in debug and produced a wrong time in release. It now uses `saturating_mul`, so the value is out of range for `Timestamp::from_microsecond`, which reports the same error jaq already produces for an out-of-range floating-point epoch.

Reproductions (before this PR, release build):

```
$ jaq -n '"ab" * 9223372036854775807'          # capacity overflow, abort
$ jaq -n '"a" * 9223372036854775807'           # allocation failure, abort
$ jaq -n '(-9223372036854775807 - 1) | length' # -9223372036854775808 (release) / panic (debug)
$ jaq -n '100000000000000000 | gmtime'         # panic (debug) / wrong time (release)
```

After this PR, each returns a normal error (or, for `length`, the correct big integer `9223372036854775808`).

Testing: full workspace test suite (including the three new regression tests), `cargo clippy -- -Dwarnings`, `cargo fmt --check`, and MSRV checks with Rust 1.69 (jaq-core) and 1.70 (workspace) all pass.

### DIFF
--- a/jaq-json/src/lib.rs
+++ b/jaq-json/src/lib.rs
@@ -627,10 +627,10 @@ impl core::ops::Mul for Val {
                 s * Num(Int(bigint_to_int_saturated(&i)))
             }
             (BStr(s), Num(Int(i))) | (Num(Int(i)), BStr(s)) if i > 0 => {
-                Ok(Self::byte_str(s.repeat(i as usize)))
+                Ok(Self::byte_str(repeat_str(&s, i as usize)?))
             }
             (TStr(s), Num(Int(i))) | (Num(Int(i)), TStr(s)) if i > 0 => {
-                Ok(Self::utf8_str(s.repeat(i as usize)))
+                Ok(Self::utf8_str(repeat_str(&s, i as usize)?))
             }
             // string multiplication with negatives or 0 results in null
             // <https://jqlang.github.io/jq/manual/#Builtinoperatorsandfunctions>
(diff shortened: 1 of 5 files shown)


01mf02:

PR #454, opened 2026-07-18
Limit nesting depth in CBOR and XML readers to prevent stack overflow

vlasky:
The CBOR and XML readers parse containers recursively without any depth check, so a small deeply nested input aborts the whole process with a stack overflow:

```
$ python3 -c "import sys; sys.stdout.buffer.write(b'\x81'*100000)" | jaq --from cbor .
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting
$ python3 -c "print('<a>'*200000)" | jaq --from xml .
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

This is also reachable through `fromcbor` / `fromxml` on untrusted strings, where a filter author would reasonably expect a catchable error rather than an abort. (The YAML reader is already safe: saphyr enforces its own nesting limit.)

**First commit**: both readers now track the nesting depth and reject documents nested deeper than 128 containers (the same limit as serde_json) with an ordinary parse error, e.g.

```
$ python3 -c "import sys; sys.stdout.buffer.write(b'\x81'*100000)" | jaq --from cbor .
Error: failed to parse: maximal nesting depth (128) exceeded
```

The choice of 128: debug builds spend several KB of stack per nesting level in the XML reader, and Rust test threads get 2 MB by default (I found that a limit of 1024 overflows there, and 256 fits only barely), so 128 keeps a comfortable margin for debug builds, small thread stacks, and wasm. Real documents nested deeper than 128 levels should be essentially nonexistent. As a side effect, the limit also makes the worst-case pre-allocation bound described at `MAX_CAP` in the CBOR reader actually enforced (previously the "8192 arrays" in that calculation was an assumption without a mechanism).

**Second commit** fixes a papercut that the new error made visible: `Tag::tag_pos` took the error position from the prefix span, which for tags without a namespace prefix is an empty span at position 0, so every error about an unprefixed tag reported `at 1:1` regardless of where the tag was. It now uses the local-name position when the prefix is empty:

```
# before
$ jaq -n '"<r>\n  <a></b></r>" | try fromxml catch .'
"... expected closing tag for a (at 1:1), found b (at 1:1)"
# after
"... expected closing tag for a (at 2:4), found b (at 2:8)"
```

Testing: five new regression tests (nesting at the limit parses, one past the limit errors, for both formats; plus one for the error positions). Full workspace test suite, `cargo clippy -- -Dwarnings`, `cargo fmt --check`, MSRV check with Rust 1.70, and a `jaq-play` check for `wasm32-unknown-unknown` all pass.

### DIFF
--- a/jaq-fmts/src/read/cbor.rs
+++ b/jaq-fmts/src/read/cbor.rs
@@ -34,11 +34,20 @@ const BUF_SIZE: usize = 4096;
 ///
 /// - 16 bytes is the size of a [`jaq_json::Val`] (as should be the case on x86_64),
 /// - 1024 is the maximal capacity that we pre-allocate, and
-/// - 8192 is the maximal stack size.
+/// - 128 is the maximal nesting depth ([`MAX_DEPTH`]).
 ///
 /// Then the maximum amount of memory that can be consumed by such an array is
-/// 16 bytes/value * 1024 values/array * 8192 arrays = 134217728 bytes = 128MB.
+/// 16 bytes/value * 1024 values/array * 128 arrays = 2097152 bytes = 2MB.
+/// (Nested maps can consume twice that amount, because
+/// every map entry holds a key and a value.)
 const MAX_CAP: usize = 1024;
+
+/// Maximal nesting depth of containers (arrays/maps).
+///
+/// Parsing a container recurses, so without this limit,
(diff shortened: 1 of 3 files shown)


01mf02:

PR #464, opened 2026-08-26
Make `no_std` work

flyingmutant:
First of all, thanks a lot for this wonderful family of libraries!

Tried to use jaq in a `no_std` context and failed (std was pulled in through dependencies), this is my attempt to make it work.

- `thumbv7em-none-eabihf` in CI is to exercise in a real `no_std` environment
- `floor`, `round` and `ceil` reimplementations to avoid an unconditional `libm` dependency (waiting for https://github.com/rust-lang/rust/issues/137578 to land!)

### Checklist

- [x] I have read the [contribution guidelines](../CONTRIBUTING.md).

### DIFF
--- a/jaq-json/Cargo.toml
+++ b/jaq-json/Cargo.toml
@@ -21,7 +21,7 @@ serde = ["serde_core"]
 jaq-core = { workspace = true }
 jaq-std  = { workspace = true }
 
-bstr = { version = "1", default-features = false, features = ["alloc"] }
+bstr = { version = "1.6", default-features = false, features = ["alloc", "unicode"] }
 bytes = { version = "1", default-features = false }
 foldhash = { version = "0.1", default-features = false }
 hifijson = { version = "0.5.0", default-features = false, features = ["alloc"] }
--- a/jaq-json/Cargo.toml
+++ b/jaq-json/Cargo.toml
@@ -21,8 +21,8 @@ serde = ["serde_core"]
 jaq-core = { workspace = true }
 jaq-std  = { workspace = true }
 
-bstr = "1"
-bytes = "1"
+bstr = { version = "1", default-features = false, features = ["alloc"] }
+bytes = { version = "1", default-features = false }
 foldhash = { version = "0.1", default-features = false }
 hifijson = { version = "0.5.0", default-features = false, features = ["alloc"] }
 indexmap = { version = "2.0", default-features = false }
--- a/jaq-json/Cargo.toml
+++ b/jaq-json/Cargo.toml
@@ -21,7 +21,7 @@ serde = ["serde_core"]
 jaq-core = { workspace = true }
 jaq-std  = { workspace = true }
 
-bstr = { version = "1.6", default-features = false, features = ["alloc", "unicode"] }
+bstr = { version = "1", default-features = false }
 bytes = { version = "1", default-features = false }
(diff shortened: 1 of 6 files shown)


01mf02:
Hi Gregory! Thanks a lot for your contribution!

Just a few days ago, I read ["How `no_std` are `no_std` Rust crates really? A survey"](https://w-graj.net/posts/rust-no-std-survey/) and wondered whether jaq was indeed `no_std`, especially concerning floating-point operations. And there you are. Did you also read the article? Or, out of curiosity, are you actually using jaq on a `no_std` architecture?

Anyway, I am on boat with your `f64` changes. However, I think I would like to keep the `ByteSlice::trim` functions the way they were before. That would make this PR a bit smaller. I think it should suffice to enable the `unicode` feature of `bstr` to make it compile on `no_std`, or am I missing something?

01mf02:
Whoops, the check fails. I am wondering why `-D warnings` seems to not have been enabled before?

... Ah, now I understand. It was enabled before, but only for the `test` and `release` workflows (because they also use `actions-rust-lang/setup-rust-toolchain@v1`), and there, no warnings appear, because we always use default features there.

Anyway, the warnings look rather easy to address. Do you want to do this?

flyingmutant:
I saw the article flying around, but didn't read it. I am using jaq parser to create on top of it a streaming version of jq subset for a `no_std` WASM environment with a custom CBOR-based data format. Sounds pretty crazy but it seems to work :-)

I'll address the warnings, sure!

Regarding the `trim` changes, `bstr` docs currently [say](https://docs.rs/bstr/latest/bstr/#cargo-features) for the `unicode` feature:

>  Note that currently, enabling this feature also requires enabling the std feature. It is expected that this limitation will be lifted at some point.

but i think `bstr` *does* compile with `unicode` and no `std` today. I'll recheck that and if that is indeed the case – revert the changes.

jaq-json/Cargo.toml
```diff
@@ -21,7 +21,7 @@ serde = ["serde_core"]
 jaq-core = { workspace = true }
 jaq-std  = { workspace = true }
 
-bstr = { version = "1", default-features = false, features = ["alloc"] }
+bstr = { version = "1.6", default-features = false, features = ["alloc", "unicode"] }
```
01mf02: Do we really need the `unicode` feature also in `jaq-json`?

jaq-std/tests/funs.rs
```diff
@@ -254,7 +254,6 @@ fn trim() {
     give(json!(" foo  "), "trim", json!("foo"));
     give(json!("  foo  "), "trim", json!("foo"));
     give(json!("foo  "), "trim", json!("foo"));
-    give(json!("\u{2003}foo\u{2003}"), "trim", json!("foo"));
```
01mf02: What about these tests? I think they would be nice to include ...

jaq-json/Cargo.toml
flyingmutant: No, my bad.

jaq-std/tests/funs.rs
flyingmutant: Will put back – removed to minimize the diff.

01mf02:
Thanks a lot again for your work and your considering my remarks! :)

